Issue #52, opened 2015-08-09
ScalaCache requires Guava Cache[String, Object]


mslinn:
Guava's CacheBuilder does not require that the resulting cache be Cache[String, Object], but as far as I can see, ScalaCache does not support any other value type for a custom CacheBuilder. Here is some sample code that compiles:

```
  import com.google.common.cache
  import com.google.common.cache._
  import java.util.concurrent.TimeUnit
  import scalacache._
  import scalacache.guava._

  val underlyingGuavaCache: cache.Cache[String, Object] = CacheBuilder.newBuilder()
    .concurrencyLevel(4)
    .softValues()
    .expireAfterWrite(5, TimeUnit.MINUTES)
    .build[String, Object]
  implicit val scalaCache = ScalaCache(GuavaCache(underlyingGuavaCache))
```

... and this is an example of what I'd like to be able to do:

```
  import com.google.common.cache
  import com.google.common.cache._
  import java.util.concurrent.TimeUnit
  import scalacache._
  import scalacache.guava._

  val underlyingGuavaCache: cache.Cache[String, String] = CacheBuilder.newBuilder()
    .concurrencyLevel(4)
    .softValues()
    .expireAfterWrite(5, TimeUnit.MINUTES)
    .build[String, String]
  implicit val scalaCache = ScalaCache(GuavaCache(underlyingGuavaCache))
```

Is there a way to accomplish this?

cb372:
No, currently there is no support for bounding the type of the cached values. It's an interesting idea, as it would allow for completely type-safe caching, but I haven't really explored it yet in ScalaCache.

The main priority for ScalaCache is to make the simple case as simple as possible, and I think that this would probably end up polluting the API with type parameters and/or member types, thereby making it more complex to use. So it would probably be best to keep the current API as-is and provide a separate, "typed" API for people who want to use it.

For now you'll just have to use a `Cache[String, Object]` even though you only want to cache strings, and turn a blind eye to the type-unsafety of it.

mslinn:
Here is a working example of how I do this, without using your package, in case this is helpful:

````
object GMap extends App {
  import scala.concurrent.ExecutionContext.Implicits._
  type Key = String
  type Value = String

  def defaultValue(key: Key): Value = s"$key value"

  val gCache = TimedCache[Key, Value]()

  val x: Value = gCache.getWithDefault("2", defaultValue("2")) // computes defaultValue
[...]

PR #55, opened 2015-08-19
Add a typed API for more typesafe caching

cb372:
Use of the typed API means you can't accidentally put something of the wrong type into your cache.

e.g.

```
import scalacache._
implicit val sc: ScalaCache = ...

val intCache = typed[Int]

intCache.put("one")(1) // OK
intCache.put("two")(2.0) // Nope! Compiler error

val one = intCache.get("1") // returns a Future[Option[Int]]
```

Use of the typed API is optional. Users can still use the existing, un-typesafe API if they prefer to play fast and loose with their types, so this change is backward compatible.

### DIFF
--- a/core/src/main/scala/scalacache/package.scala
+++ b/core/src/main/scala/scalacache/package.scala
@@ -6,6 +6,99 @@ import scala.util.{ Failure, Success, Try }
 
 package object scalacache extends StrictLogging {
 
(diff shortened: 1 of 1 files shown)


cb372:
IMO this fixes #52 but I'm open to debate about it.

PR #56, opened 2015-09-02
Fix import for memoization in README

theomega:
The import was wrong, the memoization package is in the scalacache package


cb372:
The imports in the README are correct.

```
import scalacache._
import memoization._
```

Achieves the same as

```
import scalacache.memoization._
```

(although, to be fair, relative imports are [frowned upon](http://twitter.github.io/effectivescala/#Formatting-Imports) by Twitter's Effective Scala guide).

PR #58, opened 2015-09-16
Reduce boilerplate with macro bundles

philwills:
http://docs.scala-lang.org/overviews/macros/bundles.html

### DIFF
--- a/core/src/main/scala/scalacache/memoization/Macros.scala
+++ b/core/src/main/scala/scalacache/memoization/Macros.scala
@@ -5,41 +5,37 @@ import scala.reflect.macros.blackbox
 import scala.concurrent.duration.Duration
 import scalacache.{ Flags, ScalaCache }
 
-object Macros {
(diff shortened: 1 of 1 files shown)


philwills:
Initially I was trying to add support for 2.10 via macro-compat, but the combination of a significant API change in scala-logging and having no idea how to implement `c.internal.enclosingOwner` in 2.10 made me bail on that.

cb372:
Thanks, that's much nicer!

Back in 2.10.x we had access to all kinds of useful stuff like `c.enclosingMethod`, `c.enclosingClass`, etc. But in 2.11 they decided that letting macros know too much about their surroundings was bad form and they removed all those methods. I pray that `c.internal.enclosingOwner` will stick around for a while ...

Here's how it used to work with Scala 2.10.3. https://github.com/cb372/scalacache/blob/0.1.1/core/src/main/scala/cacheable/Macros.scala#L16

PR #66, opened 2015-10-17
Add ability to use ShardedJedis for sharded caching with Redis

jareddellitt:
This adds the ability to use consistent hashing when caching to multiple instances of Redis by using [ShardedJedis](https://github.com/xetorthio/jedis/wiki/AdvancedUsage#shardedjedis).

### DIFF
--- a/redis/src/main/scala/scalacache/redis/ShardedRedisCache.scala
+++ b/redis/src/main/scala/scalacache/redis/ShardedRedisCache.scala
@@ -1,0 +1,131 @@
+package scalacache.redis
+
+import java.nio.charset.Charset
+
+import com.typesafe.scalalogging.StrictLogging
+import redis.clients.jedis._
+
+import scala.concurrent._
+import scala.concurrent.duration._
+import scalacache.{ Cache, LoggingSupport }
+
+class ShardedRedisCache(jedisPool: ShardedJedisPool, override val customClassloader: Option[ClassLoader] = None)(implicit execContext: ExecutionContext = ExecutionContext.global)
+    extends Cache
+    with RedisSerialization
+    with LoggingSupport
+    with StrictLogging {
+
+  import ShardedRedisCache.StringWithUtf8Bytes
+
+  /**
+   * Get the value corresponding to the given key from the cache
+   * @param key cache key
+   * @tparam V the type of the corresponding value
+   * @return the value, if there is one
+   */
+  override def get[V](key: String) = Future {
+    blocking {
+      withJedisClient { client =>
+        val resultBytes = Option(client.get(key.utf8bytes))
+        val result = resultBytes.map(deserialize[V])
+        logCacheHitOrMiss(key, result)
+        result
+      }
+    }
+  }
+
+  /**
+   * Insert the given key-value pair into the cache, with an optional Time To Live.
+   * @param key cache key
+   * @param value corresponding value
+   * @param ttl Time To Live
+   * @tparam V the type of the corresponding value
+   */
+  override def put[V](key: String, value: V, ttl: Option[Duration]) = Future {
+    blocking {
+      withJedisClient { client =>
+        val keyBytes = key.utf8bytes
+        val valueBytes = serialize(value)
+        ttl match {
+          case None => client.set(keyBytes, valueBytes)
+          case Some(Duration.Zero) => client.set(keyBytes, valueBytes)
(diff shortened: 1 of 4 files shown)


cb372:
Sorry about the slow reply, I just got back from holiday.

Thanks for the PR! I glanced through the code and it looks fine. I'll give it a proper review asap.

redis/src/main/scala/scalacache/redis/ShardedRedisCache.scala
```diff
@@ -0,0 +1,131 @@
...
+object ShardedRedisCache {
+
+  /**
+   * Create a sharded Redis client connecting to the given hosts and use them for caching
+   */
+  def apply(hosts: (String, Int)*): ShardedRedisCache = {
+    import scala.collection.JavaConversions.seqAsJavaList
+
+    val pool = new ShardedJedisPool(new JedisPoolConfig(), hosts.map {
+      case (host, port) ⇒ new JedisShardInfo(host, port)
```
cb372: No Unicode chars in the code please

redis/src/main/scala/scalacache/redis/ShardedRedisCache.scala
```diff
@@ -0,0 +1,131 @@
...
+   */
+  def apply(jedisPool: ShardedJedisPool, customClassloader: Option[ClassLoader] = None): ShardedRedisCache =
+    new ShardedRedisCache(jedisPool, customClassloader)
+
+  private val utf8 = Charset.forName("UTF-8")
+
+  /**
+   * Enrichment class to convert String to UTF-8 byte array
+   */
+  private implicit class StringWithUtf8Bytes(val string: String) extends AnyVal {
```
cb372: Stuff like this helper class can easily be shared with `RedisCache`. Please move it out of `RedisCache`'s companion object into a new utils object for easy sharing across the two implementations (I'd call the utils object `StringEnrichment` or something like that.

redis/src/main/scala/scalacache/redis/ShardedRedisCache.scala
```diff
@@ -0,0 +1,131 @@
...
+
+  import ShardedRedisCache.StringWithUtf8Bytes
+
+  /**
+   * Get the value corresponding to the given key from the cache
+   * @param key cache key
+   * @tparam V the type of the corresponding value
+   * @return the value, if there is one
+   */
+  override def get[V](key: String) = Future {
```
cb372: All the methods apart from `removeAll` are straight copy-pastes of the ones in `RedisCache`, and some of them are quite complex, so I'd like to share the implementations if possible. Could you refactor your code to reduce the code duplication?

Perhaps you could introduce a common trait that both `RedisCache` and `ShardedRedisCache` extend. Both `Jedis` and `ShardedJedis` implement `JedisCommands`, so you should be able to change the `withJedisClient` method to take a `f: JedisCommands => T`. (You'll also need a separate `withShardedJedisClient` method in `ShardedRedisCache` to use for sharded-specific methods like `removeAll`.)

redis/src/test/scala/scalacache/redis/ShardedRedisCacheSpec.scala
```diff
@@ -0,0 +1,132 @@
...
+
+import org.scalatest.concurrent.{ Eventually, IntegrationPatience, ScalaFutures }
+import org.scalatest.time.{ Seconds, Span }
+import org.scalatest.{ BeforeAndAfter, FlatSpec, Matchers }
+
+import scala.concurrent.Future
+import scala.concurrent.duration._
+import scala.concurrent.ExecutionContext.Implicits.global
+
+class ShardedRedisCacheSpec
```
cb372: Could you refactor this spec to share the actual tests with `RedisCacheSpec`? In other words, there should be a common trait that holds all the test code, and `RedisCacheSpec` and `ShardedRedisCacheSpec` should just take care of setting up and tearing down the cache according to their respective implementations.

cb372:
Sorry, on closer inspection I've decided not to accept this PR in its current state because it duplicates a lot of code. But it's a useful feature and I'd like to accept it if you can refactor it a bit.

Please let me know if any of my comments don't make sense, or if you need any help with refactoring.

cb372:
Thank you for the PR, but I ended up implementing this myself because I wanted to do some refactoring around Redis anyway. I've credited you in the changelog.

Issue #32, opened 2015-03-08
ClassNotFoundException when deserializing List[T] from redis


tototoshi:
I encountered ClassNotFoundException when using `memoize`.
I'm using scalacache-redis 0.5.2.

This script is minimal example to reproduce the error.

```scala
case class User(id: Int, name: String)
 
object Test {
 
  import scalacache._
  import memoization._
  import redis._
 
  implicit val scalaCache = ScalaCache(RedisCache("localhost", 6379))
 
  // Works fine when the result type is User, but throws Exception when List[User]
  def getUser(id: Int): List[User] = memoize {
    List(User(id, "Taro"))
  }
 
  def main(args: Array[String]): Unit = {
    getUser(1)
  }
 
}
```

```
scalacachedebug> r
[info] Running Test
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
[error] (run-main-3a) java.lang.ClassNotFoundException: User
java.lang.ClassNotFoundException: User
        at java.net.URLClassLoader$1.run(URLClassLoader.java:372)
        at java.net.URLClassLoader$1.run(URLClassLoader.java:361)
        at java.security.AccessController.doPrivileged(Native Method)
        at java.net.URLClassLoader.findClass(URLClassLoader.java:360)
        at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
        at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
        at java.lang.Class.forName0(Native Method)
        at java.lang.Class.forName(Class.java:344)
        at java.io.ObjectInputStream.resolveClass(ObjectInputStream.java:626)
        at java.io.ObjectInputStream.readNonProxyDesc(ObjectInputStream.java:1613)
        at java.io.ObjectInputStream.readClassDesc(ObjectInputStream.java:1518)
        at java.io.ObjectInputStream.readOrdinaryObject(ObjectInputStream.java:1774)
        at java.io.ObjectInputStream.readObject0(ObjectInputStream.java:1351)
        at java.io.ObjectInputStream.readObject(ObjectInputStream.java:371)
        at scala.collection.immutable.List$SerializationProxy.readObject(List.scala:477)
        at sun.reflect.GeneratedMethodAccessor7.invoke(Unknown Source)
[...]

Issue #35, opened 2015-03-08
Support custom serialization


cb372:
Currently hardcoded to use Java serialization for both Redis and Memcached. Might be nice to give people some options, e.g. JSON, Mgspack

ymeymann:
I managed to use JSON (specifically json4s jackson) with Redis by overriding serialize and deserialize methods on the cache. We store only one type of objects in the cache (let's call it Foo), which is a case class. The code is roughly:

```scala
val redisCache = new RedisCache(new JedisPool(host, port)) {
    override def serialize(value: Any): Array[Byte] = {
      val b = new ByteArrayOutputStream()
      serialization.write[Foo](value.asInstanceOf[Foo], b)
      b.toByteArray
    }
    override def deserialize[A](bytes: Array[Byte]): A = {
      val res = serialization.read[Foo](new String(bytes, "UTF8"))
      res.asInstanceOf[A]
    }
}
```

lloydmeta:
I think something along the lines of what Shade does with having a[ `Codec` typeclass](https://github.com/alexandru/shade/blob/master/src/main/scala/shade/memcached/Codec.scala) for `put` and `get` operations might make sense. For in-memory caches (like Guava), we can just ignore that parameter.

To make things convenient, we can have default primitive type class instances and provide an auto Java-serialisation-based one via another import. People who wish to use their own JSON or Msgpack serialisation can simply provide a single conversion to `Codec`. e.g. if using Play JSON

```scala
implicit def toScalacacheCodec(implicit javaFormat: Format[A]): Codec[A] = new Codec[A] {
  def serialize(o: A): Array[Bytes] = {
    val json = o.toJson
    json.toString.getBytes("UTF-8")
  }
  def deserialize(data: Array[Bytes]): A = {
    val json = Json.parse(new String(data, "UTF-8"))
    json.as[A]
  }
}
```

joprice:
:+1: I've had to give up using this for remote caching for this reason.

cb372:
@joprice You'll be happy to hear that ScalaCache 0.8.0 adds support for custom serialization.

Closing this issue.

joprice:
And on my birthday! Thank you!

ymeymann:
Thanks @cb372 & @lloydmeta, and happy birthday @joprice! Just upgraded to 0.8.0, and it is working.

lloydmeta:
Whoa. Happy birthday @joprice :)

PR #86, opened 2016-03-20
Refs #35: Custom generic serialisation via type class

lloydmeta:
Just a prototype of [what I mentioned on the issue](https://github.com/cb372/scalacache/issues/35#issuecomment-198845958), to serve as a point of discussion and garner feedback.

Goals:
- Provides sensible serialisation for primitive types out of the box
- If desired, users can import from an object housing a Java-serialising object serialiser
- If end users have their own object-to-MsgPack/JSON/whatever codecs, they just need to provide a suitable auto converter to our `Codec` (conversely, this project can have projects that provide bindings for commonly used formats)

Changes:
- Add Codec typeclass to allow adhoc customisation of serialisation/deserialisation
- Add some default Codecs for standard primitives w/ tests
- Add a Java-basd object serialiser w/ test
- Modify various existing methods, macros, and tests to make this work

Misc:
- Bump PGP plugin version


cb372:
Sorry for the slow review.

I just spent ages writing a huge, detailed comment but it got deleted. Let's see if I can summon the willpower to write it all out again. The sacrifices we make for FOSS...

I agree that a typeclass is a good way to implement custom serialisation. I've experimented with it in the past, but never quite got around to implementing it properly. This PR looks like a great start. Just a couple of general comments before I go through the code and make a few OCD-driven requests for changes.

## Compatibility

Changing the serialisation format is a big breaking change for Memcached and Redis users. After upgrading ScalaCache their cache will be effectively cold and their logs will probably start filling up with warnings about deserialisation failures.

In one of my previous experiments with custom serialisation, I tried implementing a `Codec` that emulated the behaviour of the default Spymemcached `Transcoder` in order to preserve compatibility. But it turned out to be quite tricky because the data would have to pass through both our codec and Spy's transcoder, in effect becoming double-serialised.

I think a more pragmatic solution would be to add a boolean flag to `MemcachedCache` and `RedisCache` that disables the codec-based serialisation and uses the existing implementation. This would at least let people decide if and when they want to take the hit of the serialisation format change. This flag would also need to be clearly documented in the README and the changelog.

## Compression

Spymemcached's [default transcoder](https://github.com/couchbase/spymemcached/blob/master/src/main/java/net/spy/memcached/transcoders/BaseSerializingTranscoder.java) automatically gzips anything over 16KB. Might be nice to do that in ScalaCache as well, at least as an option. But it does complicate things, as you then need to write a header containing a compression flag.

## Import tax

I'd like to keep imports to a minimum. Ideally just writing `import scalacache._` should bring the Java serialisation codecs into scope as low priority implicits. I think we can achieve this by making the `scalacache` package object extend from `JavaSerializationCodecs`. Can a package object extend from a trait? Not sure, probably yes.

caffeine/src/main/scala/scalacache/caffeine/CaffeineCache.scala
```diff
@@ -1,5 +1,6 @@
 package scalacache.caffeine
 
+import scalacache.serdes.Codec
```
cb372: Please can we make the package `scalacache.serialization`?

core/src/main/scala/scalacache/package.scala
```diff
@@ -126,7 +128,7 @@ package object scalacache extends StrictLogging {
    *
    * @tparam V the type of values that the cache will accept
    */
-  def typed[V](implicit scalaCache: ScalaCache) = new TypedApi[V]()(scalaCache)
+  def typed[V: Codec](implicit scalaCache: ScalaCache) = new TypedApi[V]()(scalaCache, implicitly[Codec[V]])
```
cb372: Please add a parameter to the implicit param list instead of using implicitly.

(See slide 84 of http://plastic-idolatry.com/erik/sw2015.pdf)

memcached/src/main/scala/scalacache/memcached/MemcachedCache.scala
```diff
@@ -19,16 +21,18 @@ class MemcachedCache(client: MemcachedClient, keySanitizer: MemcachedKeySanitize
...
    */
-  override def get[V](key: String) = {
+  override def get[V: Codec](key: String) = {
     val p = Promise[Option[V]]()
     val f = client.asyncGet(keySanitizer.toValidMemcachedKey(key))
     f.addListener(new GetCompletionListener {
       def onComplete(g: GetFuture[_]): Unit = p.complete {
-        val result = Option(f.get.asInstanceOf[V])
+        val byteArray = Option(f.get.asInstanceOf[Array[Byte]])
+        val result = byteArray.map(implicitly[Codec[V]].deserialize)
```
cb372: Same comment as above about `implicitly`

redis/src/main/scala/scalacache/redis/RedisCacheBase.scala
```diff
@@ -45,15 +47,16 @@ trait RedisCacheBase
...
    * @tparam V the type of the corresponding value
    * @return the value, if there is one
    */
-  final override def get[V](key: String) = Future {
+  final override def get[V: Codec](key: String) = Future {
     blocking {
       withJedisCommands { jedis =>
         val resultBytes = Option(jedis.get(key.utf8bytes))
-        val result = resultBytes.map(deserialize[V])
+        val result = resultBytes.map(implicitly[Codec[V]].deserialize)
```
cb372: implicitly

core/src/main/scala/scalacache/serdes/BaseCodecs.scala
```diff
@@ -0,0 +1,112 @@
+package scalacache.serdes
+
+trait BaseCodecs {
+
```
cb372: Where did this serialisation code come from? If it's lifted from somewhere (Shade?) then we should at least give credit in a code comment.

core/src/main/scala/scalacache/serdes/JavaSerializationCodec.scala
```diff
@@ -0,0 +1,72 @@
...
+
+/**
+  * Object input stream which tries the thread local class loader.
+  *
+  * Thread Local class loader is used by SBT to avoid polluting system class loader when
+  * running different tasks.
+  *
+  * This allows deserialization of classes from sub-projects during something like
+  * Play's test/run modes.
+  */
```
cb372: Interesting. Maybe this could have solved #32. Or, maybe it will cause a regression in the fix for that bug...

The fix was to allow users to supply their own classloader to `RedisCache`, but that will now be ignored.

Needs a bit more investigation.

core/src/test/scala/scalacache/PackageObjectSpec.scala
```diff
@@ -20,25 +20,25 @@ class PackageObjectSpec extends FlatSpec with Matchers with BeforeAndAfter with
   behavior of "#get"
 
   it should "call get on the cache found in the ScalaCache" in {
-    scalacache.get("foo")
+    scalacache.get[String]("foo")
```
cb372: Does adding the `: Codec` cause type inference to stop working as expected? 

e.g. if this was `val foo: String = scalacache.get("foo")` would it still compile without a type annotation? If not, that's kind of a shame, but not the end of the world.

Somebody somewhere probably has some code that will need a type annotation to be added, so we should put a warning in the changelog.

caffeine/src/main/scala/scalacache/caffeine/CaffeineCache.scala
lloydmeta: Cool done.

core/src/main/scala/scalacache/package.scala
lloydmeta: Interesting.  didn't know there was a cost to `implicitly`. Fixed.

memcached/src/main/scala/scalacache/memcached/MemcachedCache.scala
lloydmeta: Fixed.

redis/src/main/scala/scalacache/redis/RedisCacheBase.scala
lloydmeta: Fixed.

core/src/main/scala/scalacache/serdes/BaseCodecs.scala
lloydmeta: Added a comment to give credit to Shade (which I maintain now BTW).

core/src/main/scala/scalacache/serdes/JavaSerializationCodec.scala
lloydmeta: For what it's worth, the tests you added when you resolved #32 still pass (`Issue32Spec` and `PlayIntegrationSpec`).

core/src/test/scala/scalacache/PackageObjectSpec.scala
lloydmeta: Yeah, unfortunately, it looks like having the implicitly resolved `Codec[A]` (doesn't seem to matter if it is done via a context bound or separate implicit param) seems to require us to state the type param when calling the method. Added a note to the changelog.

lloydmeta:
> I just spent ages writing a huge, detailed comment but it got deleted. Let's see if I can summon the willpower to write it all out again. The sacrifices we make for FOSS...

:cry: I hate it when that happens. Thanks for rewriting your feedback; I know it isn't easy.

> Compatibility

Good point; I've added a `useLegacySerialization` parameter to the Memcached and Redis cache implementations that bypasses usage of the `Codec` when set to true.

> Compression

I think that is a great idea. Would it be ok with you if that were to be added on in another PR ? Since we would be compressing and decompressing bytes, it seems reasonable to treat compression as a separate concern.

> Import tax

I hear you on the wanting to lower the amount of things that people need to import, but `scalacache._` seems like the main bootstrapping import for reasonably kosher things that are relatively stable.

Mixing in `JavaSerializationCodecs ` into that package object would mean that Java-serialisation-based Codec would be available by default.  Since Java serialisation is really slow, I think it might be better to have it be available only after paying a small import tax, as a warning to developers using the lib. It seems better to save `scalacache._` for when we come up with some really nice and fast generic `Codec` derivation-magic so that we can avoid deserialisation problems at that point in time. It also saves us from having an `Any` Codec in-scope inside that package, which is useful in case we forget to declare implicit params when adding/modifying the code.

Also, when using generic typeclasses (whether it's deriving them or bringing things in scope that derive them), I think people are familiar with having to do an additional import; Play, Argonaut, Circe etc have somewhat established this pattern.

cb372:
## Compression

Yep, let's add it in another PR. Could even be a separate codec implementation.

## Import tax

I think Java serialisation is a sensible default, even if it is slow. It's good enough for a lot of use cases (and hasn't put many people off using ScalaCache until now). Also, low barrier to entry is a high priority for me - a lot of people don't want to care how their stuff is being serialised, they just want to cache stuff as easily as possible.

And don't forget about people using Guava/Caffeine/Ehcache/Lru. They would have to import a serialisation codec that they won't even use, which is both annoying and confusing.

lloydmeta:
> Import tax

Ok got it. I think your reasoning is sound. Done.

CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
+0.7.6
+----
+
+New features:
+
+* [#86](https://github.com/cb372/scalacache/pull/86) adds support for custom serialisation. Thanks to @lloydmeta
```
cb372: Haha, thanking yourself in the changelog. Nice :)

lloydmeta: That is some deep existential stuff right there.

Or just blind copy pasting of existing stuff. :laughing:

PR #89, opened 2016-04-04
Implement compression

lloydmeta:
Adds support for compression/decompression of serialised data. 

Current implementation is as a mixin to any `Codec` and uses GZip as you mentioned in #86. Another approach would be to add a field on caches that send data over the wire like Redis or Memcached that does something similar.

Changes:
- Refactor some old code to allow for extension
- Add a GzippingCodec that can be used to add Gzipping on any Codec
- Add tests for CompressionJavaCodec
- Misc refactoring

### DIFF
--- a/README.md
+++ b/README.md
@@ -290,8 +290,18 @@ types, and provide an implementation for objects based on Java serialisation.
 
 ### Custom Codec
 
-If you want to use a custom `Codec` for your object of type `A`, simply implement an instance of `Codec[A]` and make sure it
+If you want to use a custom `Codec` for your object of type `A`, simply implement an instance of `Codec[A, Array[Byte]]` and make sure it
 is in scope at your `set`/`put` call site.
+
+### Compression of `Codec[A, Array[Byte]]`
+
+If you want to compress your serialised data before sending it to your cache, ScalaCache has a built-in `GZippingBinaryCodec[A]` mix-in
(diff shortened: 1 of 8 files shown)


joprice:
:+1: Another great addition!

cb372:
This looks great, but I'll hold off on merging it until I see how #91 pans out.

cb372:
Sorry, this now needs a rebase. Once you've fixed it up I'll merge it and release 8.1.

lloydmeta:
Just rebased. Let me know if you want anything changed.

lloydmeta:
[The failure](https://travis-ci.org/cb372/scalacache/builds/122159248#L1409) looks....confusing/scary

> ArrayBuffer((scalacache.memoization.SquerylIntegrationSpec.findUser(1),Some(scalacache.memoization.User@1),None)) had size 0 instead of expected size 1 (SquerylIntegrationSpec.scala:58)

The `ArrayBuffer` being described by the assertion error clearly has a size of 1 😖 

I think this is because `memoizeSync` eventually awaits `_caching`, and that method has a fire-and-forget [`asynchronouslyCacheResult`](https://github.com/cb372/scalacache/blob/master/core/src/main/scala/scalacache/package.scala#L56) call: it returns the result without waiting for the cache `put` to finish. This makes sense to me, but it means that cache put may not have been called yet immediately after returning from `memoizeSync`. To fix this test, I've added an `eventually { .. }` in ccd7357.

cb372:
Makes sense. Surprising that it hasn't failed until now. Thanks for fixing.

PR #91, opened 2016-04-07
Add a dummy codec for non-Serializable classes

### LINKED ISSUE #90: Codec needed to cache (Seq of) custom types in Caffeine cache
After upgrading from 0.7.5 to 0.8.0 of scalacache, compilation of

```
import scalacache._
import caffeine._

implicit val scalaCache = ScalaCache(CaffeineCache())

def baz(): Seq[foo.Bar] = ...

def logic(): [foo.Bar] = sync.cachingWithTTL(key)(ttl)(baz)
```

fails at cachingWithTTL call with

> Could not find any Codecs for type Seq[foo.Bar]. Please provide one or import scalacache._

Caffeine is in-memory cache - why is codec i.e. serialization/deserialization needed at all?

cb372:
Fixes #90

The tradeoff is that it's now possible to attempt to serialise a non-Serializable object using Redis/Memcached and get a runtime exception. But that was also true with the legacy serialisation.

### DIFF
--- a/caffeine/src/test/scala/foo/Issue90Spec.scala
+++ b/caffeine/src/test/scala/foo/Issue90Spec.scala
@@ -1,0 +1,30 @@
+package foo
+
+import org.scalatest._
+
+/**
+ * Replicating the issue reported in https://github.com/cb372/scalacache/issues/90
+ */
+class Issue90Spec extends FlatSpec with Matchers {
+
+  "Issue 90" should "be fixed" in {
+    import scalacache._
+    import caffeine._
+    import scala.concurrent.duration._
+
+    implicit val scalaCache = ScalaCache(CaffeineCache())
+
+    def baz(): Seq[foo.Bar] = Seq(
+      Bar(1, "one"),
+      Bar(2, "two")
+    )
+    val ttl = 1.second
+
+    """
+    def logic(): Seq[foo.Bar] = sync.cachingWithTTL("the-key")(ttl)(baz())
+    """ should compile
+  }
+
+}
+
+case class Bar(a: Int, b: String)
--- a/core/src/main/scala/scalacache/serialization/JavaSerializationCodec.scala
+++ b/core/src/main/scala/scalacache/serialization/JavaSerializationCodec.scala
@@ -6,11 +6,34 @@ import scala.reflect.ClassTag
 import scala.util.control.NonFatal
 
 /**
+  * The world's least useful codec.
+  * Just throws a runtime exception if you try to use it.
+  *
+  * Only exists so that we can provide codecs for any class <: AnyRef,
+  * so that people don't run into "codec missing" issues when using in-memory cache impls e.g. Caffeine.
+  *
+  * See https://github.com/cb372/scalacache/issues/90
+  */
+trait AnyRefDummyLowPriorityCodec {
+
+  implicit def AnyRefBinaryCodec[S <: AnyRef](implicit classTag: ClassTag[S]) = new Codec[S] {
+
+    def serialize(value: S): Array[Byte] =
+      sys.error(s"Sorry, I don't know how to serialize an instance of class ${classTag.runtimeClass.getName}. Please provide a custom Codec.")
+
+    def deserialize(data: Array[Byte]): S =
+      sys.error(s"Sorry, I don't know how to deserialize an instance of class ${classTag.runtimeClass.getName}. Please provide a custom Codec.")
+
+  }
+
+}
+
+/**
   * Holds a Java-serialisation-based Codec[Object <: Serializable] instance
   *
   * Credit: Shade @ https://github.com/alexandru/shade/blob/master/src/main/scala/shade/memcached/Codec.scala
   */
-trait JavaSerializationCodec {
+trait JavaSerializationCodec extends AnyRefDummyLowPriorityCodec {
 
   private[this] class GenericCodec[S <: Serializable](classTag: ClassTag[S]) extends Codec[S] {
 
@@ -43,7 +66,7 @@ trait JavaSerializationCodec {
   /**
     * Uses plain Java serialization to deserialize objects
     */
-  implicit def AnyRefBinaryCodec[S <: Serializable](implicit ev: ClassTag[S]): Codec[S] =
+  implicit def SerializableBinaryCodec[S <: Serializable](implicit ev: ClassTag[S]): Codec[S] =
     new GenericCodec[S](ev)
 
 }


joprice:
What about another mixin with an implicit that makes this implicit ambiguous in case you want to statically avoid this behavior? Or extracting the TypedApi into a trait that the package object here mixes in, so you can avoid the fallback codec altogether?

cb372:
@joprice 

>What about another mixin with an implicit that makes this implicit ambiguous in case you want to statically avoid this behavior?

I like this idea, but it will require some investigation. I'll have to read up on implicit resolution order to work out how to implement this.

I didn't really understand the second half of your comment. Could you elaborate?

joprice:
The `JavaSerializationCodec` is used by this package object: https://github.com/cb372/scalacache/blob/07282f61e45d1ea018efe69e30f03168ec7abbd3/core/src/main/scala/scalacache/package.scala#L9

This ties `JavaSerializationCodec` together with `TypedApi`. Instead, I'm imagining either `JavaSerializationCodec` or `TypedApi` on its own, so that the implicits of `JavaSerializationCodec` aren't imported by default. So maybe there would be two imports: 

```scala
object JavaSerializationCodec extends  JavaSerializationCodec
import scalacache._
import JavaSerializationCodec._
```

Or you could have the default or compose your own:

```scala
trait Typed {
  class TypedApi[V](implicit val scalaCache: ScalaCache, codec: Codec[V]) {
...
}

package object scalacache extends StrictLogging with Typed

object scalacache extends StrictLogging with Typed
```

lloydmeta:
Hallo,

 I think #90 can be broken down into 2 issues that are loosely related:

1. In-memory implementations of `scalacache.Cache` still need suitable `Codec`s at the cache call site, even though the `Codec` isn't actually used.
2. We don't provide `Codec`s for commonly used standard-lib base traits, such as `Seq` and `Map`

### Regarding 1

I think as long as ScalaCache is committed to abstracting away the distinction between in-memory and remote caches by providing a single interface to do caching with (really the root cause of @sslavic's woes), we should provide some kind of unsafe `Any` Codec to make life easier for people using in-memory caches. 

The `AnyRefBinaryCodec` in this PR adds such a codec, but I share @joprice's reservations on whether such a codec should be available by default (`import scalacache._` is essentially default), since it circumvents the type system, opening the door to potentially nasty runtime errors for people who are using it with Redis, Memcached, or any cache that calls for serialisation. Basically, it comes down to finding a good compromise between preventing potential run-time errors for people who do need `Codec`, and avoiding unnecessary compile-time errors for people who don't need `Codec`, and I think having this codec in-scope by default swings too much in the latter direction.

Perhaps a good compromise is to have this new unsafe codec made available under an aptly-named import (e.g. `import scalacache.serialization.InMemoryCodec._`)? Sure, that adds a small import tax on in-memory cache users to solve their compile-time woes, but it helps prevent unpredictable *run-time* behaviour for remote cache users.

### Regarding 2

I think we should probably add Codecs for commonly used collection base traits that do not extend `Serializable`. At the moment, I can only think of `Seq`, and `Map`. Though `Map` [can be tricky](http://stackoverflow.com/questions/32900862/map-can-not-be-serializable-in-scala), we can have an implicit that provides `Codec[Seq[A]]`s that convert `Seq`s to `Vector`s before using standard Java serialisation on it.

We could potentially just keep adding new default `Codec`s (this is the fate of using typeclasses in general) when more complaints come in, and perhaps eventually add a nice `generic-codec` sub library that is backed by a nice serialising lib (maybe this should be on the roadmap).

### Summary

I think solving issue 2) resolves #90 as well, but in the long run, #1 needs to be addressed :)

cb372:
Unfortunately I didn't have any success with the ambiguous implicits idea. The implicit specificity rules just don't work the way we want them to. For some reason `A <: Serializable` and `A <: AnyRef` are treated as having the same specificity, so it's very tricky (impossible?) to provide an ambiguous implicit for "AnyRef but not Serializable".

However, I have sketched out an alternative solution that I think works quite nicely. See [this gist](https://gist.github.com/cb372/235754f3a9e6372101906b3023a6d7db). I don't have time to flesh it out into a full PR right now (I'm supposed to be writing my Scala Days talk!), but I think it'll work.

The idea is to add a type param `Repr` to `Codec` saying what representation the codec converts things into, e.g. `Array[Byte]`. Then we allow each `Cache` implementation to specify, again with a type param, what representation they want to work with. For Memcached/Redis it would be `Array[Byte]`; for in-memory caches it's a dummy type meaning they don't require any serialization.

The only downside I can foresee is that the `Repr` type param leaks into the API for `get`, `caching`, etc., so users might have to write things like

```scala
val myString = get[String, Array[Byte]](key)
```

But I don't think that's a show-stopper. I still think this is better than any of the alternatives we've come up with so far.

lloydmeta:
I see; that's unfortunate.

From your gist:

> The idea is to add a type param Repr to Codec saying what representation the codec converts things into, e.g. Array[Byte]. Then we allow each Cache implementation to specify, again with a type param, what representation they want to work with

I think this makes sense; I have a (non-working...) prototype that was going in this direction as well. To be clear, the only place where there is leakage of `Repr` in your example is where someone wants to wrap calls to `Cache` in a method like [here](https://gist.github.com/cb372/235754f3a9e6372101906b3023a6d7db#file-gistfile1-txt-L59). If you just want to get or put values into a cache, the API still [looks quite clean](https://gist.github.com/cb372/235754f3a9e6372101906b3023a6d7db#file-gistfile1-txt-L60), no?

cb372:
>the only place where there is leakage of Repr in your example is where someone wants to wrap calls to Cache in a method

Yes, that's correct, but that's exactly what we do in the package object.

>This still leaves us without Codecs for commonly-used abstract collection types, which should be added for people who are using remote caches.

Yep, agreed.

lloydmeta:
Since you don't have time cuz of friggin Scaladays (not jealous at all....no), would you like me try to set up the PR?

lloydmeta:
Nevermind. Fixed a bunch of compile errors and threw it up for review.

joprice:
@cb372  Have you tried bringing out the big guns? https://github.com/milessabin/shapeless/blob/b73453bab27e73b6aa1206805e73ec583852309d/core/src/ main/scala/shapeless/lowpriority.scala#L6

cb372:
No, I didn't try that, although it looks awesome/terrifying. I didn't want to go any further down the rabbit hole with this, as I wasn't entirely happy with this "dangerous by default, but you can add a mixin to make it safe again" solution anyway.

cb372:
Closing. Merged #92 in preference to this PR.

PR #92, opened 2016-04-09
#90 and #91: Add Repr to Codec

lloydmeta:
A quick implementation of [this idea](https://github.com/cb372/scalacache/pull/91#issuecomment-207808528)

- Extend Codec to have a Repr parameter that represents the end result of performing serialisation with the codec
- Make Cache implementations specify the format in which they store data (corresponds to the kind of Repr
  they need), for example In-memory caches like Guava will specify InMemoryRepr
- Add a no-op Codec[Any, InMemoryRepr] codec

### DIFF
--- a/caffeine/src/main/scala/scalacache/caffeine/CaffeineCache.scala
+++ b/caffeine/src/main/scala/scalacache/caffeine/CaffeineCache.scala
@@ -1,12 +1,13 @@
 package scalacache.caffeine
 
-import scalacache.serialization.Codec
-import scalacache.{ LoggingSupport, Cache, Entry }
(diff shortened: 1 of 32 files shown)


cb372:
Thanks for doing the drudge work! I bet that was boring.

Do you want to cherry-pick https://github.com/cb372/scalacache/pull/91/commits/5fe83af380bfe463c0f44f486b7ab6e5faf28c55 onto this branch, just so we have a direct test for the reported issue?

lloydmeta:
No worries, it was quick.

> Do you want to cherry-pick 5fe83af onto this branch, just so we have a direct test for the reported issue?

Oh, good idea. Done :)

PR #102, opened 2016-05-21
Optimisations for get and memoize

cb372:
Optimization of `get`, `caching` and `memoize`

* Hand-optimise `CacheKeyBuilder` (used by both `get` and `memoize`) to use a loop and a StringBuilder
* Hand-optimise `MethodCallToStringConverter` (used by `memoize`) similarly
* Special treatment for `memoize` because we know in this case that the key is a single string, not an `Any*`
* Special handling in `_caching` for already-completed futures (such as those returned by the in-memory cache implementations). If the future is already completed, we can avoid calling `recover` and `flatMap` on it, which means we avoid spawning more futures. This makes a HUGE difference. This one optimisation accounts for the vast majority of the speedup.

## Before

```
[info] Benchmark                                      Mode  Cnt      Score      Error  Units
[info] CaffeineBenchmark.caffeineGet                  avgt   30     20.547 ±    0.596  ns/op
[info] CaffeineBenchmark.scalacacheGetNoMemoize       avgt   30    334.866 ±   32.630  ns/op
[info] CaffeineBenchmark.scalacacheGetTypedNoMemoize  avgt   30    340.825 ±   36.284  ns/op
[info] CaffeineBenchmark.scalacacheGetWithMemoize     avgt   30  39801.345 ± 2974.877  ns/op
```

## After

```
[info] Benchmark                                      Mode  Cnt    Score     Error  Units
[info] CaffeineBenchmark.caffeineGet                  avgt   30   20.110 ±   0.270  ns/op
[info] CaffeineBenchmark.scalacacheGetNoMemoize       avgt   30  198.882 ±  10.764  ns/op
[info] CaffeineBenchmark.scalacacheGetTypedNoMemoize  avgt   30  206.160 ±  12.344  ns/op
[info] CaffeineBenchmark.scalacacheGetWithMemoize     avgt   30  580.810 ± 129.729  ns/op
```

Not bad for an afternoon's work 😄  

/cc @wuservices @ben-manes in case you're interested

### DIFF
--- a/core/src/main/scala/scalacache/CacheKeyBuilder.scala
+++ b/core/src/main/scala/scalacache/CacheKeyBuilder.scala
@@ -7,6 +7,11 @@ trait CacheKeyBuilder {
    */
   def toCacheKey(parts: Seq[Any])(implicit cacheConfig: CacheConfig): String
 
+  /**
+   * Build a cache key by prepending the configured prefix, if there is one
+   */
+  def stringToCacheKey(key: String)(implicit cacheConfig: CacheConfig): String
+
 }
 
 object DefaultCacheKeyBuilder extends CacheKeyBuilder {
@@ -16,8 +21,47 @@ object DefaultCacheKeyBuilder extends CacheKeyBuilder {
    * Prepends the prefix if there is one.
    */
   override def toCacheKey(parts: Seq[Any])(implicit cacheConfig: CacheConfig): String = {
-    val prefix = cacheConfig.keyPrefix.map(p => s"$p${cacheConfig.keySeparator}") getOrElse ""
-    s"$prefix${parts.mkString(cacheConfig.keySeparator)}"
+    // Implementation note: the type of `parts` will be `WrappedArray` when called from the package object,
+    // so random access is O(1).
+    val sb = new StringBuilder(128)
+    val separator = cacheConfig.keySeparator
+
+    // Add the key prefix if there is one
+    cacheConfig.keyPrefix match {
+      case Some(prefix) =>
+        sb.append(prefix)
+        sb.append(separator)
+      case None => // do nothing
+    }
+
+    var i = 0
+
+    // Add all key parts except the last one, with the separator after each one
+    while (i < parts.size - 1) {
+      sb.append(parts(i))
+      sb.append(separator)
+      i += 1
+    }
+    // Add the final key part
+    if (i < parts.size) {
+      sb.append(parts(i))
+    }
+
+    sb.toString
+  }
+
+  override def stringToCacheKey(key: String)(implicit cacheConfig: CacheConfig): String = {
+    cacheConfig.keyPrefix match {
+      case Some(prefix) =>
+        val separator = cacheConfig.keySeparator
+        val sb = new StringBuilder(prefix.length + separator.length + key.length)
+        sb.append(prefix)
+        sb.append(separator)
+        sb.append(key)
+        sb.toString
+      case None =>
+        key // just return the key as is
+    }
   }
 
 }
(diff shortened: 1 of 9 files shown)


core/src/main/scala/scalacache/CacheKeyBuilder.scala
```diff
@@ -16,8 +21,47 @@ object DefaultCacheKeyBuilder extends CacheKeyBuilder {
    * Prepends the prefix if there is one.
    */
   override def toCacheKey(parts: Seq[Any])(implicit cacheConfig: CacheConfig): String = {
```
ben-manes: What about a fast path for a singleton collection (1 part)?

core/src/main/scala/scalacache/package.scala
```diff
@@ -50,21 +59,38 @@ package object scalacache extends JavaSerializationCodec {
...
-        case None =>
-          val result: Future[From] = f
-          asynchronouslyCacheResult(result)
-          result
+      val fromCache: Future[Option[From]] = getWithKey(key)
+
+      if (fromCache.isCompleted) {
+        // optimisation for in-memory caches that return Future.successful(...)
+        fromCache.value.get match {
+          case Success(Some(value)) => Future.successful(value)
```
ben-manes: It seems unfortunate that every read creates a wrapper, especially when some caches support memoization directly. Perhaps you could push the `calculateAndCacheResult` into the non-memoizing caches so that the returned future is bubbled directly to the caller?

core/src/main/scala/scalacache/CacheKeyBuilder.scala
ben-manes: Oh, I think `stringToCacheKey` is basically what I was looking for!

core/src/main/scala/scalacache/CacheKeyBuilder.scala
```diff
@@ -16,8 +21,47 @@ object DefaultCacheKeyBuilder extends CacheKeyBuilder {
...
+
+  override def stringToCacheKey(key: String)(implicit cacheConfig: CacheConfig): String = {
+    cacheConfig.keyPrefix match {
+      case Some(prefix) =>
+        val separator = cacheConfig.keySeparator
+        val sb = new StringBuilder(prefix.length + separator.length + key.length)
+        sb.append(prefix)
+        sb.append(separator)
+        sb.append(key)
+        sb.toString
```
ben-manes: I'd expect a little faster by avoiding an intermediate array copy if you used `prefixWithSeperator.concat(key)`. I'd expect you could probably precompute the prefixWithSeperator and use it elsewhere too

ben-manes: Benchmarked. Seems the JVM optimized StringBuilder making it faster.

cb372: Thanks for checking. I didn't think of using `concat`.

core/src/main/scala/scalacache/package.scala
cb372: That's a good idea. I'll try that in a separate PR after merging this.

cb372:
@ben-manes thanks for the review!

core/src/main/scala/scalacache/package.scala
ben-manes: Saw this idea didn't make it into the release. If you make a success garbage free it then I'd expect similar performance to `scalacacheGetNoMemoize`.

cb372: I had a quick stab at it but got caught out by the fact that Caffeine does things synchronously whereas we are dealing with Futures in ScalaCache. I would've had to wrap a Future in an `Await.result()`, which is always a bad sign. 

I'd definitely like to revisit this, but it may require a non-trivial redesign of the cache API to more accurately capture the differences between sync and a sync impls. In hindsight, making Caffeine/Guava share exactly the same API as Caffeine/Redis is maybe a bit restrictive.

ben-manes: Caffeine has an asynchronous option, but Guava doesn't. But why can't the cached value be a future?

Issue #100, opened 2016-05-21
Optimise all the things


cb372:
As we discovered in #82, ScalaCache has quite a high overhead compared to accessing the underlying cache directly. Let's see what we can do to improve this.

mdedetrich:
I have noticed that you have recently removed scala-logging as a dependency. If you are using slf4j directly, then it will have worse performance than scala-logging. The main reason is that scala-logging (and other logging libraries) usually use an inbuilt macro that will only log if a certain logging level is enabled, i.e. if your log level is `INFO` and you do `logger.debug(...)`, it won't even call `logger.debug`.

If you still want to use slf4j directly, then it may be useful to implement a macro so that calls to `logCacheHitOrMiss` only happen if appropriate logging level is enabled

cb372:
What the scala-logging macro did was rewrite code like this:

```scala
logger.info("yolo")
```

into code like this:

```scala
if (logger.isInfoEnabled) {
  logger.info("yolo")
}
```

When I removed scala-logging, I just went through the codebase and manually rewrote all the logging calls to look like the latter. It's a bit boilerplate-y, but ScalaCache doesn't do logging in that many places. It wasn't painful enough to make me want to write a macro.

However, you're right that it would be nice to avoid the call to `logCacheHitOrMiss` if the logger is not enabled at warn level. (The method is too big for HotSpot to inline for us.) Could do this manually or with a macro.

mdedetrich:
> However, you're right that it would be nice to avoid the call to logCacheHitOrMiss if the logger is not enabled at warn level. (The method is too big for HotSpot to inline for us.) Could do this manually or with a macro.

I can do this in #103 If you wish

PR #103, opened 2016-05-22
Work on performance improvements

mdedetrich:
Ongoing work for #100

### DIFF
--- a/ehcache/src/main/scala/scalacache/ehcache/EhcacheCache.scala
+++ b/ehcache/src/main/scala/scalacache/ehcache/EhcacheCache.scala
@@ -28,17 +28,14 @@ class EhcacheCache(underlying: Ehcache)
    * @return the value, if there is one
    */
   override def get[V](key: String)(implicit codec: Codec[V, InMemoryRepr]) = {
-
-    val valueObj = underlying.get(key)
-
-    if (valueObj != null) {
-      val result = Some(valueObj.getObjectValue.asInstanceOf[V])
-      logCacheHitOrMiss(key, result)
-      Future.successful(result)
-    } else {
-      logCacheHitOrMiss(key, None)
-      Future.successful(None)
+    val result = {
+      val elem = underlying.get(key)
+      if (elem == null) None
+      else Option(elem.getObjectValue.asInstanceOf[V])
     }
+
+    logCacheHitOrMiss(key, result)
+    Future.successful(result)
   }
 
   /**
--- a/ehcache/src/main/scala/scalacache/ehcache/EhcacheCache.scala
+++ b/ehcache/src/main/scala/scalacache/ehcache/EhcacheCache.scala
@@ -28,12 +28,17 @@ class EhcacheCache(underlying: Ehcache)
    * @return the value, if there is one
    */
   override def get[V](key: String)(implicit codec: Codec[V, InMemoryRepr]) = {
-    val result = for {
-      e <- Option(underlying.get(key))
-      v <- Option(e.getObjectValue.asInstanceOf[V])
-    } yield v
-    logCacheHitOrMiss(key, result)
-    Future.successful(result)
+
+    val valueObj = underlying.get(key)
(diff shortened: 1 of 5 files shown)


cb372:
Thanks! But why are you doing if-null checks in some places and `try { ... } catch NPE` in others? Could we standardise on the former please. I think throwing and catching an exception is much slower than a null check.

mdedetrich:
> But why are you doing if-null checks in some places and try { ... } catch NPE in others? 

According to the documentation for that specific library it throws a `NullPointerException`. It may be a case of me being stupid, is throwing a `NullPointerException` the same thing as returning `null` in Java land?

cb372:
Neither Guava nor Caffeine will throw an NPE if the value is not present.

* [Javadoc for Guava's getIfPresent method](https://google.github.io/guava/releases/18.0/api/docs/com/google/common/cache/Cache.html#getIfPresent(java.lang.Object))
* [And for Caffeine's getIfPresent](http://static.javadoc.io/com.github.ben-manes.caffeine/caffeine/2.3.0/com/github/benmanes/caffeine/cache/Cache.html#getIfPresent-java.lang.Object-)

If the value is not present, both libraries will return null. So I'd recommend the following code for Guava and Caffeine:

```scala
val entry = underlying.getIfPresent(key).asInstanceOf[Entry[V]]
if (entry == null || entry.isExpired) None
else Some(entry.value)
```

(Note that calling `asInstanceOf[...]` on `null` is perfectly fine, it won't throw an NPE.)

ehcache/src/main/scala/scalacache/ehcache/EhcacheCache.scala
```diff
@@ -28,12 +28,17 @@ class EhcacheCache(underlying: Ehcache)
...
-    val result = for {
-      e <- Option(underlying.get(key))
-      v <- Option(e.getObjectValue.asInstanceOf[V])
-    } yield v
-    logCacheHitOrMiss(key, result)
-    Future.successful(result)
+
+    val valueObj = underlying.get(key)
+
+    if (valueObj != null) {
```
cb372: EhCache is a bit tricky. There are multiple cases to handle here:

* `underlying.get(key)` returns `null`
* It returns a non-null element, but `elem.getObjectValue` is `null`

I'd suggest this:

```scala
val result = {
  val elem = underlying.get(key)
  if (elem == null) None
  else Option(elem.getObjectValue.asInstanceOf[V])
}
logCacheHitOrMiss(key, result)
Future.successful(result)
```

mdedetrich:
Hmm, I swear that the docs said it would throw a NullPointerException, let me look.

For the Caffeine Cache, the documentation for `getIfPresent` states this

```java
  /**
   * Returns the value associated with {@code key} in this cache, or {@code null} if there is no
   * cached value for {@code key}.
   *
   * @param key key whose associated value is to be returned
   * @return the value to which the specified key is mapped, or {@code null} if this map contains no
   *         mapping for the key
   * @throws NullPointerException if the specified key is null
   */
  @CheckForNull
  V getIfPresent(@Nonnull Object key);
```

So according to the docs it is meant to throw null?

I will look at the other changes when I get home tonight

cb372:
To summarise the Javadoc you posted above:

* If there is nothing in the cache with the given key, it will *return* `null`
* If you pass it `null` as the key that you want to look up, it will *throw* a `NullPointerException`

We need to handle the first case by doing a null-check on the returned value.

We don't really care about the second case: If you try to look something up with a `null` key, you deserve a NullPointerException! 😛

mdedetrich:
Ah thanks I misread the docs 😜 Will do the requested adjustments today!

mdedetrich:
Did the requested changes, anything else that needs to be done?

cb372:
Thanks, this looks fine now. I ran the micro-benchmark against this branch and it looks like there is a slight speed boost thanks to the reduction in object allocations 👍

PR #106, opened 2016-05-31
logCacheHitOrMiss flag has now been placed at the call site

mdedetrich:
This just moves the `logger.isDebugEnabled` at the call site, rather than within the `logCacheHitOrMiss` function call.

I think this should improve performance, should probably verify with benchmarks


cb372:
Wow, sorry to leave this lingering for 4 months!

PR #114, opened 2016-09-26
Add Aerospike support as cache

richiesgr:
Hi
I've added the support for aerospike as it's a popular solution for caching
please give me your inputs 
Thanks

### DIFF
--- a/aerospike/src/main/scala/scalacache/aerospike/AerospikeCache.scala
+++ b/aerospike/src/main/scala/scalacache/aerospike/AerospikeCache.scala
@@ -20,7 +20,7 @@ class AerospikeCache(hostList: Array[Host],
 
   private val policy = new AsyncClientPolicy()
   policy.threadPool = ExecutionContextExecutorServiceBridge(executionContext)
-  implicit val client = new AsyncClient(policy, hostList :_*)
+  implicit val client = new AsyncClient(policy, hostList: _*)
 
   logger.info(s"Connection status to Aerospike : ${client.isConnected}")
   logger.info(s"Current Nodes in cluster : ${client.getNodes}")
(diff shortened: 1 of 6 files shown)


aerospike/src/main/scala/scalacache/aerospike/AerospikeCache.scala
```diff
@@ -0,0 +1,94 @@
...
+import org.slf4j.LoggerFactory
+
+import scala.concurrent.duration.Duration
+import scala.concurrent.{ ExecutionContext, Future }
+import scalacache.serialization.Codec
+import scalacache.{ Cache, LoggingSupport }
+import scalaz.NonEmptyList
+
+/**
+ * Created by Richard Grossman on 2016/7/18.
```
cb372: Could you delete the file headers please?

project/Build.scala
```diff
@@ -87,6 +87,15 @@ object ScalaCacheBuild extends Build {
     )
     .dependsOn(core)
 
+  lazy val aerospike = Project(id = "scalacache-aerospike", base = file("aerospike"))
+    .settings(implProjectSettings: _*)
+    .settings(
+      libraryDependencies ++= Seq(
+        "com.github.otrimegistro" %% "aerospikez" % "0.2"
```
cb372: I'm not really happy with adding this lib as a dependency. It hasn't been touched in over 2 years and it depends on a very old version of Scalaz Stream, and it has a lot of transitive dependencies. 

Adding dependencies on unmaintained Scala libs makes it difficult to release ScalaCache when new versions of Scala come out (Scala 2.12 is just around the corner! 🎉 ), and I like to keep transitive dependencies to a minimum, for the sake of ScalaCache's users.

Could you use the [official Aerospike Java client](http://www.aerospike.com/docs/client/java)?

richiesgr:
Switch to java client from aerospike

cb372:
Hmm, it looks like Travis CI doesn't have support for Aerospike out of the box. Do you want to play around with travis.yml and see how difficult it would be to install and run Aerospike on Travis?

PR #115, opened 2016-10-31
Cross-build core for Scala.js

mdedetrich:


mdedetrich:
Doesn't work for Scala 2.10.6, will figure it out today

cb372:
Looks like a good start! Are you planning to do anything else on this PR?

mdedetrich:
Well we can do two things from here.

We can check that the current pull request works in regards to releasing/publishing and merge it if it is fine. I don't think there is an issue, but it would be good to check.

Or I can try and add an implementation of an LRU cache specifically designed and optimized for Javascript. Scala-cache core has a memoization implementation that already works with Scala.js, but I suspect its performance is going to be less than ideal because it involves a lot of casting back and from `Byte` (which is pointless and irrelevant in Javascript).

I think the former is a better idea as adding tuned Scala.js implementation will take some time (and someone else may be able to provide a better implementation). With the core of scalacache in Scala.js, it will let other people add their implementations (for example with cookies and whatnot).

build.sbt
```diff
@@ -12,22 +14,31 @@ lazy val root = Project(id = "scalacache",base = file("."))
   .settings(commonSettings: _*)
   .settings(sonatypeSettings: _*)
   .settings(publishArtifact := false)
-  .aggregate(core, guava, memcached, ehcache, redis, caffeine)
+  .aggregate(coreRoot, guava, memcached, ehcache, redis, caffeine)
+
+lazy val coreRoot = Project(id = "scalacache-core", base = file("."))
```
cb372: I don't think we want this `coreRoot` project. You end up with two projects in the same directory (`"."`), which confuses sbt.

The root project should aggregate both `coreJVM` and `coreJS`, making the `coreRoot` project unnecessary.

mdedetrich: Ok done

mdedetrich:
Alrighty, I have rebased on upstream/scala-2.12.0. Let me know if there are any problems

> Are you talking about serialization?

Yes I was. I was just saying that if there is a default implementation that uses `Array[Byte]`, its probably going to have worse performance in Scala.js. In Scala.js the most performant implementation would probably be storing either Javascript Objects or Javascript values in something like a map, but thats something I need to look into

.travis.yml
```diff
@@ -1,7 +1,7 @@
 sudo: false
 language: scala
 scala:
-  - 2.11.8
+  - 2.12.0
```
cb372: Hmm, now I'm confused. I've merged in my 2.12 PR, so this line should no longer be in the diff.

Sorry, do you mind doing another rebase against master? I think GitHub is a bit messed up.

mdedetrich: That might have been my bad, let me try again

mdedetrich: Actually my bad, forgot to properly push. It should be fine nowl

cb372:
Thanks for the PR!

PR #116, opened 2016-11-02
Returns combined Future[From] results on cache

lucamilanesio:
Combine cache putWithKey result with the original
body result futures with flatMap, allowing the caller to be aware
of when the data gets actually cached and assure consistency of
expected behaviour.

Without this code change, some paradox situations were actually
possible where data was supposed to be cached but it wasn't.

Example of a failing test:

```
    val results = for {
      res1 <- cache.caching("key")(Future { "initial value" })
      res2 <- cache.caching("key")(Future { "updated value" })
    } yield (Seq(res1, res2))

    results map {
      _ should contain only ("initial value")
    }
```

### DIFF
--- a/core/src/main/scala/scalacache/package.scala
+++ b/core/src/main/scala/scalacache/package.scala
@@ -49,20 +49,22 @@ package object scalacache extends JavaSerializationCodec {
 
     private def _caching(key: String)(ttl: Option[Duration])(f: => Future[From])(implicit flags: Flags, execContext: ExecutionContext): Future[From] = {
 
-      def asynchronouslyCacheResult(result: Future[From]): Unit = result onSuccess {
+      def asynchronouslyCacheResult(result: Future[From]): Future[From] = result flatMap {
         case computedValue =>
-          Try(putWithKey(key, computedValue, ttl)) recover {
-            case e =>
+          try {
+            putWithKey(key, computedValue, ttl)
+          } catch {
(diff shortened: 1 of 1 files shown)


lucamilanesio:
@cb372 do you have a second to have a look on this PR? The problem should be simple to identify in the code and the fix is very small 😄 

Thank you in advance for your time.

Kind Regards.

cb372:
Hi Luca, thanks for the PR.

Doing the caching asynchronously is actually by design, so that you can get your result back as quickly as possible without waiting for the cache write. But a few different people have complained that it is confusing behaviour, so I think it should be configurable.

Please give me a couple of days to think about it.

lucamilanesio:
Hi @cb372, I see your point. However, a lot of times the user would like to know when the cache has been written and perform subsequent actions based on that.

We use ScalaCache and we like it a lot, because it allows us to avoid making expensive calls to our backend system. Knowing when the cache is "ready" and making those calls on a map() of the cache future will definitely help us avoiding further expensive calls.

What we see today is that just after the future is completed, the cache isn't because of the lack of the combination of the two futures.

Let me submit a further commit where, instead of combining them, I can return both futures as tuple so that the caller can decide where to combine them with a Futures.sequence or not.

cb372:
I'm closing this, as I went with a slightly different implementation of the same idea. Thank you anyway for the contribution!

PR #117, opened 2016-11-07
Scala 2.12.0

xuwei-k:


cb372:
Thanks for the PR, but I'm going to merge mine because it also bumps sbt.

I don't think there's much value in running multiple Travis builds with different Scala versions. I'll just bump Travis's Scala version to 2.12.

Issue #118, opened 2016-11-07
Documentation: Flags.readsEnabled invalidates the cache


k4200:
Hi,

I've been using Scalacache for some time. Thank you for the great library.

I faced a situation where our system needs to invalidate the cached result of a memoized method. I forked the code and was going to implement it, but then noticed that `Flags.readsEnabled = false` does that. (Correct me if I'm wrong). I thought it just skips cache reads temporarily, and doesn't affect the cached result.

Let me explain my misunderstanding using the `getUser` example for clarification.

```scala
saveToDb(User(1, "Prince"))
val user1a = getUser(1, false) // From the DB
val user1b = getUser(1, false) // Cached result
saveToDb(User(1, "The Artist ...."))
val user1c = getUser(1, true) // From the DB
val user1d = getUser(1, false) // Cached result
```

What I thought was:
* user1a = user1b = user1d = Prince
* user1c =The Artist ...

The real behavior (as far as I see from the code) is:
* user1a = user1b = Prince
* user1c = user1d = The Artist ...

Maybe because I'm not an English speaker, the sentence "Cache GETs and/or PUTs can be temporarily disabled using flags" made me misunderstand the behavior. It would be helpful for someone like me to add a little note about this.

By the way, I tested it by adding the following code to `Issue42Spec`:
```scala
  "memoize with readsEnabled" should "invalidate cache" in {
    val user1a = getUser(1)
    val user1b = getUser(1)
    val user1c = {
      implicit val flags = Flags(readsEnabled = false)
      getUser(1)
    }
    val user1d = getUser(1)
    user1a should not be user1c
    user1a should be(user1b)
    user1c should be(user1d)
  }
```

This test case fails 8 out of 10 times on my machine. A sort of timing issue?

Thank you.

k4200:
> This test case fails 8 out of 10 times on my machine. A sort of timing issue?

This is probably related to #107

cb372:
Thank you for the thorough report!

Yes, the documentation could probably use some work here. Just to confirm the current behaviour, which I think is reasonable, is as follows:

* If `readsEnabled = false`, the cache will not be read, and ScalaCache will behave as if it was a cache miss. This means that memoization will read the value from the DB and then write it to the DB.
* If `writesEnabled = false`, in the case of a cache miss, the value will be read from the DB but it will not be written to the cache.
* If both flags are false, memoization will not read from the cache or write to the cache.

I'll try to clarify the README.

[...]

PR #120, opened 2016-11-07
Enables to invalidate cache

k4200:
Add `Flags.invalidateEnabled` and change the cache behavior when it's set to true.

I'm not sure how to write tests for this. Writing to cache in tests seems to be an in memory operation, and end immediately, so reproducing the asynchronous issue might require a bit more than adding a simple test case.

Related issue: https://github.com/cb372/scalacache/issues/118

### DIFF
--- a/core/src/main/scala/scalacache/Flags.scala
+++ b/core/src/main/scala/scalacache/Flags.scala
@@ -2,12 +2,18 @@ package scalacache
 
 /**
  * Configuration flags for conditionally altering the behaviour of ScalaCache.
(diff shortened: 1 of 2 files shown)


k4200:
Just noticed that #116 deals with the same issue.

k4200:
#122 supersedes this PR.

kell18:
Hi! Thanks for the nice lib!
Does it make it to any release? I cannot see this `invalidateEnabled` in version 0.28.0

Issue #107, opened 2016-06-14
sync.caching is not deterministic


jkugiya:
When getting value from cache, it seems there is possibility that cannot be getting value from cache.
For example

```
import scalacache.sync

def getValue(key: String): String = sync.caching(key) { ... }

val a = getValue("key1")
val b = getValue("key1") // this code may not get value from cache.
```

So I read [cache's code](https://github.com/cb372/scalacache/blob/60b3b790093b88127b448de8b7c43385f6cd855c/core/src/main/scala/scalacache/package.scala#L64). Apparenly, putting value to cache is done asynchronously. So getting value from cache is not deterministic. Is this intended behavior?

jkugiya:
Following test results is not stable.

```
  import org.mockito.{ Mockito, Matchers => m }
  import scalacache._
  class Foo {
    def reverse(str: String) = str.reverse
  }
  class Bar(foo: Foo) {
    implicit val gc = ScalaCache(GuavaCache())
    def get(key: String): String = sync.caching(key) {
      foo.reverse(key)
    }
  }
  feature("sample") {
    for (i <- 1 to 100) {
      scenario(s"${i}:") {
        val foo = Mockito.mock(classOf[Foo])
        val bar = new Bar(foo)
        bar.get("test!!!")
        bar.get("test!!!")
        bar.get("test!!!")
        Mockito
          .verify(foo, Mockito.times(1))
          .reverse(m.eq("test!!!"))
      }
    }
  }
```

jendakol:
Do I get right this is already solved issue? If so, close it, please..

jkugiya:
Sorry for slow response.

PR #122, opened 2016-11-07
Make the write semantics of `caching` configurable

cb372:
By default, do not complete the `Future` until the cache write has completed. This makes the behaviour deterministic, as you have a guarantee that the value is now present in the cache when the Future completes.

Note that this is a change in behaviour. Until now, ScalaCache would complete the Future as quickly as possible once the value had been computed, and perform the cache write in the background. It is possible to maintain this behaviour by toggling a boolean in the config.


cb372:
I consider this to be a fix for issues #107 and for the non-deterministic behaviour noted in #118. It is also a duplicate of/supersedes PRs #116 and #120.

PR #127, opened 2017-06-01
Move `isDebugEnabled` check inside `logCacheHitOrMiss`

### LINKED ISSUE #126: logCacheHitOrMiss is just being called when log level is debug
So, I've spent the latest 5 hours of my life trying to figure it out why my production service was logging nothing related to the scalacache :P.

I have a custom `CaffeineCacheProvider` where I override the `logCacheHitOrMiss` so I can create stats for miss/hit 
```scala
  protected override def logCacheHitOrMiss[A](key: String, result: Option[A]): Unit = {
    mystuff
    super.logCacheHitOrMiss(key, result)
}
```
Then everything worked perfectly on my local machine, dev and staging environment, but we couldn't see logs/stats on production.

I found out that that the lib is not calling the `logCacheHitOrMiss` method if log library debug is not enabled 🤕 

https://github.com/cb372/scalacache/blob/955528c379356918509c126b8e7365a0bbdb44e5/caffeine/src/main/scala/scalacache/caffeine/CaffeineCache.scala#L46
image

cb372:
Fixes #126

### DIFF
--- a/caffeine/src/main/scala/scalacache/caffeine/CaffeineCache.scala
+++ b/caffeine/src/main/scala/scalacache/caffeine/CaffeineCache.scala
@@ -43,8 +43,7 @@ class CaffeineCache(underlying: CCache[String, Object])
         if (entry.isExpired) None else Some(entry.value)
       } else None
     }
-    if (logger.isDebugEnabled)
-      logCacheHitOrMiss(key, result)
+    logCacheHitOrMiss(key, result)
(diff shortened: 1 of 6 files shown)


LATaylor-guardian:
👍 (for old times' sake)

cb372:
Cheers Luke!

diego-alvarez-hs:
Thank you @cb372

PR #143, opened 2017-09-30
Make the return types more generic

hussachai:
I use scalacache interface for my custom cache implementation. It abstracts away a lot of boilerplate code. Great work! 
I found small things in the Cache trait that I believe it could be improved, and I'd love to know what you think about it.
```scala
trait Cache[Repr] {
  ...
  def put[V](key: String, value: V, ttl: Option[Duration])(implicit codec: Codec[V, Repr]): Future[Unit]
  def remove(key: String): Future[Unit]
  def removeAll(): Future[Unit]
}
```

The return type of put, remove, and removeAll are Future[Unit]. Semantically, Unit is similar to void in Java and they are compatible. The problem is a lot of cache implementations do not return Unit type (or void). They often return Boolean, Integer, or Long.
I implemented the scalacache for rediscala and the following code may illustrate the problem better than my words.

```scala
class MyRediscalaCache(...) {
  import system.dispatcher
 ...
override def put[V](key: String, value: V, ttl: Option[Duration])(implicit codec: Codec[V, Array[Byte]]): Future[Unit] = {
    val valueBytes = codec.serialize(value)
    RedisCache.put(redisClient, key, valueBytes, ttl).map(_ ⇒ ())
  }
  override def remove(key: String): Future[Unit] = {
    RedisCache.remove(redisClient, key).map(_ ⇒ ())
  }
  override def removeAll(): Future[Unit] = {
    RedisCache.removeAll(redisClient).map(_ ⇒ ())
  }
  ...
}
```
When the return type is Future[Unit], I could not return Future[Boolean] or Future[Long]. 
I have to `map` one type to another and it requires ExecutionContext in the scope.
While the computation is cheap, the `map` function uses Promise and ExecutionContext to execute the conversion in a separated thread. If we change the return type of those methods to Future[Any], we can don't have to do `map(_ => ())`.

I don't know how people use this Cache, but this change should not break most people code (unless they create an abstract over an abstract. still in that case, it is an easy fix for them) 
What do you think?

### DIFF
--- a/modules/core/shared/src/main/scala/scalacache/Cache.scala
+++ b/modules/core/shared/src/main/scala/scalacache/Cache.scala
@@ -23,7 +23,7 @@ trait Cache[Repr] {
    * @param ttl Time To Live
    * @tparam V the type of the corresponding value
    */
-  def put[V](key: String, value: V, ttl: Option[Duration])(implicit codec: Codec[V, Repr]): Future[Unit]
+  def put[V](key: String, value: V, ttl: Option[Duration])(implicit codec: Codec[V, Repr]): Future[Any]
 
   /**
    * Remove the given key and its associated value from the cache, if it exists.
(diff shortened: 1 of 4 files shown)


cb372:
Thanks for the contribution! 

That looks fine to me. I very much doubt that anybody is relying on the return value of these methods.

Technically it is a breaking change though, so I'll bump the version to 0.10 when I release this.

hussachai:
@cb372 Awesome. Thank you.

PR #156, opened 2018-01-05
Add jvmopts file to set max memory

jleider:
Fixes OOM error when running tests when running sbt locally. It appears that travis has a default Xmx setting of 2G.


jleider:
Travis failed cloning the repo.. I dont have access to rebuild.

PR #173, opened 2018-04-16
Add an implicit conversion for cats-effect Async

### LINKED ISSUE #172: Finally tagless effect and Free of IO
This is a proposal to remove the `Mode[IO]` in favor of a more generic approach using the implicit conversion to the type class `cats.effect.Effect`.

### The problem

In my projects I use the finally tagless approach to Free my code from side-effect and build a architecture around `Effect[F]` without compromising to any monadic instance – hence `IO` in this context but literally any effect type class – so I can interchange seamlessly to any class that has a conversion to `Effect`.

Whereas the current implementation of Cats mode provided by Scalacache sets that I cannot derive from `Effect` directly even though `IO` has a instance of it:

```scala
import cats.effect.Effect
import scalacache.CatsEffect.mode._
import scalacache._
import scalacache.memoization._

trait Users[F[_]] {
  def findById(id: Long): F[Option[User]]
}

object Users {
  implicit def usersEffect[F[_]](implicit F: Effect) = new Users[F] {
    // This does not work out of the box since there is no implicit Mode[F]
    def findById(id: Long) = memoizeF[F, Option[User]](None)(???)
  }

  implicit def usersIO = new Users[IO] {
    // Compromised to the IO monad to all eternity :(
    def findById(id: Long) = memoizeF[IO, Option[User]](None)(???)
  }
}
```

This means in practice that I have to wrap every class with `IO` from the beginning, event though the implementation does not require it.

### The solution

We could leverage the type class implicit conversion where the `IO`:

```scala
import scalacache.CatsEffect.asyncForCatsEffectAsync
import scalacache.{Async => AsyncCache, Mode}

implicit def cacheModeEffect[F[_]](implicit F: Effect[F]) = new Mode[F]
   val M: AsyncCache[F] = asyncForCatsEffectAsync(F)
}
``` 

Actually all the required implementation is already provided by the `asyncForCatsEffectAsync`.

This approach has a second benefit since to integrate to virtually any other library that uses cats under the hood out of the box – including [Monix](https://github.com/monix/monix/blob/master/monix-eval/shared/src/main/scala/monix/eval/internal/TaskConversions.scala#L73) – that provide an implicit conversion for `Effect[F]`.

Nonetheless,  the conversion is not too far from your `Mode` type implementation but just a common pattern around effectful monads that is used by several Typelevel projects to extend libs functionality (is [monix.tail.Iterant](https://monix.io/api/3.0/monix/tail/Iterant.html)).

### Finally

Despite the full report I don't have too much experience with Scalacache and the design decisions around the project.

Let me know If you find this proposal pertinent and the ways I can contribute.

Thanks for the good work!

cb372:
Fixes #172

### DIFF
--- a/modules/cats-effect/src/main/scala/scalacache/CatsEffect.scala
+++ b/modules/cats-effect/src/main/scala/scalacache/CatsEffect.scala
@@ -16,6 +16,14 @@ object CatsEffect {
       val M: Async[IO] = asyncForCatsEffectAsync[IO]
     }
 
+    /**
+      * A mode that wraps computations in F[_],
+      * where there is an instance of cats-effect Async available for F.
+      */
+    implicit def async[F[_]](implicit F: CatsAsync[F]): Mode[F] = new Mode[F] {
+      val M: Async[F] = asyncForCatsEffectAsync[F]
+    }
+
   }
 
   def asyncForCatsEffectAsync[F[_]](implicit af: CatsAsync[F]): Async[F] = new Async[F] {


modules/cats-effect/src/main/scala/scalacache/CatsEffect.scala
```diff
@@ -16,6 +16,14 @@ object CatsEffect {
       val M: Async[IO] = asyncForCatsEffectAsync[IO]
```
hilios: IMO the old `Mode[IO]` should be removed in favor of the Effectul type class.

But this would impact users that do not have the `import cats.effect.implicits._` in the scope.

PR #174, opened 2018-04-16
Add test coverage analysis and bump coverage

mrvisser:
Added scoverage for coverage analysis. We can run `sbt clean coverage test` to run the tests, then `sbt coverageReport` to generate the reports.

I've disabled it for ScalaJS as it causes the build to fail.


cb372:
Thanks! LGTM

PR #175, opened 2018-04-25
Add support for OHC

### LINKED ISSUE #161: OHC support
https://github.com/snazy/ohc

Current GCs of Oracle/Open JDK JVMs cannot handle cache workload efficiently for caches with high write rate, so off-heap maps/caches only tools that usable.

rider-yi:
Fixes #161

### DIFF
--- a/modules/benchmarks/src/main/scala/scalacache/benchmark/OhcBenchmark.scala
+++ b/modules/benchmarks/src/main/scala/scalacache/benchmark/OhcBenchmark.scala
@@ -1,0 +1,64 @@
+package scalacache.benchmark
+
+import java.util.concurrent.TimeUnit
+
+import org.caffinitas.ohc.OHCacheBuilder
+import org.openjdk.jmh.annotations._
+import org.openjdk.jmh.infra.Blackhole
+
+import scalacache._
+import scalacache.memoization._
+import scalacache.modes.sync._
+import scalacache.ohc.OhcCache
+
+@State(Scope.Thread)
+class OhcBenchmark {
+
+  val underlyingCache =
+    OHCacheBuilder
+      .newBuilder()
+      .keySerializer(OhcCache.stringSerializer)
+      .valueSerializer(OhcCache.stringSerializer)
+      .timeouts(true)
+      .build()
+  implicit val cache: Cache[String] = OhcCache(underlyingCache)
+
+  val key = "key"
+  val value: String = "value"
+
+  def itemCachedNoMemoize(key: String): Id[Option[String]] = {
+    cache.get(key)
+  }
+
+  def itemCachedMemoize(key: String): String = memoizeSync(None) {
+    value
+  }
+
+  // populate the cache
+  cache.put(key)(value)
+
+  @Benchmark
+  @BenchmarkMode(Array(Mode.AverageTime))
+  @OutputTimeUnit(TimeUnit.NANOSECONDS)
+  def scalacacheGetNoMemoize(bh: Blackhole) = {
+    bh.consume(itemCachedNoMemoize(key))
+  }
+
+  @Benchmark
+  @BenchmarkMode(Array(Mode.AverageTime))
+  @OutputTimeUnit(TimeUnit.NANOSECONDS)
+  def scalacacheGetWithMemoize(bh: Blackhole) = {
+    bh.consume(itemCachedMemoize(key))
+  }
+
+  @Benchmark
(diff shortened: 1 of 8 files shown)


modules/ohc/src/test/scala/scalacache/ohc/OhcCacheSpec.scala
```diff
@@ -0,0 +1,99 @@
...
+
+  import scalacache.modes.sync._
+
+  behavior of "get"
+
+  it should "return the value stored in the underlying cache" in {
+    val underlying = newOHCache
+    underlying.put("key1", "hello")
+    OhcCache(underlying).get("key1") should be(Some("hello"))
+    underlying.close()
```
rider-yi: I think we have to close OHC caches explicitly because OHC allocates off-heap memory.

modules/benchmarks/src/main/scala/scalacache/benchmark/OhcBenchmark.scala
```diff
@@ -0,0 +1,64 @@
...
+  @BenchmarkMode(Array(Mode.AverageTime))
+  @OutputTimeUnit(TimeUnit.NANOSECONDS)
+  def scalacacheGetWithMemoize(bh: Blackhole) = {
+    bh.consume(itemCachedMemoize(key))
+  }
+
+  @Benchmark
+  @BenchmarkMode(Array(Mode.AverageTime))
+  @OutputTimeUnit(TimeUnit.NANOSECONDS)
+  def cache2kGet(bh: Blackhole) = {
```
cb372: This method is misnamed

modules/benchmarks/src/main/scala/scalacache/benchmark/OhcBenchmark.scala
```diff
@@ -0,0 +1,64 @@
...
+    bh.consume(underlyingCache.get(key))
+  }
+
+  @TearDown
+  def close(): Unit = cache.close()
+
```
cb372: Did you run the benchmarks? Is OHC actually fast?

modules/ohc/src/test/scala/scalacache/ohc/OhcCacheSpec.scala
cb372: That makes sense

modules/benchmarks/src/main/scala/scalacache/benchmark/OhcBenchmark.scala
fan-yu-ishikawa: Yes, I did and the result is this:
```
sbt:benchmarks> jmh:run -i 5 -wi 5 -f1 -t1 .*OhcBenchmark.*
[info] # Run complete. Total time: 00:00:42
[info] Benchmark                              Mode  Cnt    Score    Error  Units
[info] OhcBenchmark.ohcGet                    avgt    5  171.455 ±  1.786  ns/op
[info] OhcBenchmark.scalacacheGetNoMemoize    avgt    5  209.608 ± 16.892  ns/op
[info] OhcBenchmark.scalacacheGetWithMemoize  avgt    5  422.980 ± 11.269  ns/op
```

As a matter of fact, OHC is 3-15 times slower than Caffeine because off-heap cache has serialization/deserialization overhead when getting/setting objects.
But I think OHC can be used to avoid GC pressure when using a very huge number of objects in a very large memory.

rider-yi: Yes, I did and the result is this:
```
sbt:benchmarks> jmh:run -i 5 -wi 5 -f1 -t1 .*OhcBenchmark.*
[info] # Run complete. Total time: 00:00:42
[info] Benchmark                              Mode  Cnt    Score    Error  Units
[info] OhcBenchmark.ohcGet                    avgt    5  171.455 ±  1.786  ns/op
[info] OhcBenchmark.scalacacheGetNoMemoize    avgt    5  209.608 ± 16.892  ns/op
[info] OhcBenchmark.scalacacheGetWithMemoize  avgt    5  422.980 ± 11.269  ns/op
```

As a matter of fact, OHC is 3-15 times slower than Caffeine because off-heap cache has serialization/deserialization overhead when getting/setting objects.
But I think OHC can be used to avoid GC pressure when using a very huge number of objects in a very large memory.

cb372:
Sorry, would you mind rebasing? You conflicted with yourself 😄 

Don't worry about the failing Travis build. It's flaky occasionally.

PR #179, opened 2018-05-11
Update libs

guizmaii:
- sbt to v1.1.5
 - cats-effect to v1.0.0-RC1

Complementary to https://github.com/cb372/scalacache/pull/178


cb372:
I'm not sure about depending on an RC version of cats-effect. Maybe just wait until 1.0 comes out?

Do you know whether it's binary-compatible with 0.10?

PR #185, opened 2018-05-27
[WIP] Scalacache should be polymorphic on the types he saves

guizmaii:
As discussed on Gitter with @mdedetrich, I think that Scalacache could be easier to use if `Cache` was not parameterize with the value `V`.

I'm trying to change that here.

For now, It just compiles. There're still some work to do. For example, for now, there're some ugly `Any` that I should remove.


mdedetrich:
Thanks for the work, I assume this is only going to make it for the next major release?

guizmaii:
@mdedetrich did you read the @cb372 answer he gave us in Gitter ? I'm not sure that he is interested by my implementation. 

I still have to answer to him but I have to think more about my answer. 
@cb372 have good arguments and a good vision of what its lib should be. We should respect that. It's his work.

Nevertheless, I'll continue my work here because there're still tests not passing and I think I could address some @cb372 remarks but I have to test my ideas.

mdedetrich:
Yeah I saw the answer, you can probably do the best of both worlds, i.e, make the base cache constructor polymorphic on its methods, but provide another constructor (via a cache) which forces the type parameter to be constant within the internal methods

guizmaii:
There's a lot of work to do and I do it only on my spare time. I'll need one good month to finish it, IMO.

tbrown1979:
Ahh, I commented after you made some changes it seems. The comments may still be relevant for the change to `Async`

guizmaii:
Hi @tbrown1979,

Thanks for the comments but this PR is still very WIP. It's maybe a bit too early for a review. 

Can I ping you when my PR is ready for review ?

tbrown1979:
@guizmaii Sorry about that! Yeah, a ping when it's ready would be great 😄I should have noticed the [WIP] in the title..

guizmaii:
I don't have to time to code that.

PR #186, opened 2018-06-09
Added basic cache stats

perotom:
Would be great to have basic stats for caches. Especially helpful for tests which want to verify if data was retrieved from the cache or not.

TODO: Add documentation


guizmaii:
IMO, a better idea is just to make the `underlying` cache accessible. This `stats` operation is not something share between all cache implementations. Scalacache is abstraction of common cache operations and should not address specific use cases like this one.

perotom:
Also no problem. Is it possible to make the underlying cache accessible in the next release? Heavily depend on it for testing

cb372:
I agree, just turning the `underlying` parameter in the `CaffeineCache` constructor into a `val` is probably a simpler solution. I don't see any problem with doing that.

lewisjkl:
Looks like this has been resolved and then left alone for quite some time. I am going to close, but we can reopen if anyone desires.

PR #190, opened 2018-07-22
Cross-compile most modules against Scala 2.13.0-M4

cb372:
Everything that doesn't depend on cats can support 2.13.0-M4.

There was only one breakage caused by the new collections library: `scala.IndexedSeq` is now a type alias for `scala.collections.immutable.IndexedSeq`, so we can no longer use an ArrayBuffer (see the change in Macros.scala).


modules/core/shared/src/test/scala/scalacache/memoization/MemoizeSpec.scala
```diff
@@ -265,7 +265,7 @@ class MemoizeSpec extends FlatSpec with Matchers with ScalaFutures with Eventual
   class MockDbCall(result: String) extends (Int => String) {
     val calledWithArgs = ArrayBuffer.empty[Int]
     def apply(a: Int): String = {
-      calledWithArgs.append(a)
+      calledWithArgs += a
```
cb372: `append` is deprecated in the new collections

project/plugins.sbt
```diff
@@ -3,7 +3,7 @@ addSbtPlugin("com.47deg"          % "sbt-microsites"        % "0.7.17")
 addSbtPlugin("com.jsuereth"       % "sbt-pgp"               % "1.1.0")
 addSbtPlugin("com.github.gseitz"  % "sbt-release"           % "1.0.7")
 addSbtPlugin("com.lucidchart"     % "sbt-scalafmt-coursier" % "1.15")
-addSbtPlugin("org.scala-js"       % "sbt-scalajs"           % "0.6.22")
+addSbtPlugin("org.scala-js"       % "sbt-scalajs"           % "0.6.24")
```
cb372: TODO: this causes the plugin to emit some deprecation warnings during the build. Will sort that out in a separate PR.

lewisjkl:
Going to close this since Scala 2.13 is already supported within the project.

PR #197, opened 2018-09-06
cats-effect 1.0.0, Monix 3.0.0-RC2

cb372:
Monix 3.0.0-RC2 depends on cats-effect 1.0.0-RC3, which is pretty damn close to 1.0.0.

I wonder if Monix 3.0.0 will be released soon? We could wait for that.

### DIFF
--- a/.jvmopts
+++ b/.jvmopts
@@ -1,1 +1,2 @@
--Xmx1G
+-Xms4G
+-Xmx4G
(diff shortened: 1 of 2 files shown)


dborisenko:
Hi @cb372, is it possible to release scalacache only for cats-effect 1.0.0 without Monix? We don't use Monix, but we use both `cats-effect` and `scalacache`. And this PR blocks us from doing an upgrade to the latest version of `cats-effect`

cb372:
Just released this is v0.25.0. It's on it's way to Maven Central.

dborisenko:
Great! Thank you, @cb372 . And thank you for your awesome library 👍

ploddi:
Why Monix version is `3.0.0-RC2-d0feeba` and not just `3.0.0-RC2`?

PR #220, opened 2018-11-09
Adding helper function to instanciate caches within Cats Resource

GrafBlutwurst:
I couldn't figure out where and what sort of meaningful tests to add. so I'd be happy for any pointers.

### DIFF
--- a/modules/cats-effect/src/main/scala/scalacache/CatsEffect.scala
+++ b/modules/cats-effect/src/main/scala/scalacache/CatsEffect.scala
@@ -1,6 +1,7 @@
 package scalacache
 
 import cats.Functor
+import cats.evidence.<~<
 import cats.effect.{Async => CatsAsync, IO, Resource}
 
 import scala.language.higherKinds
@@ -52,9 +53,8 @@ object CatsEffect {
 
   }
 
-  final case class ResourceCache[F[_]: CatsAsync: Functor: Mode]() {
-    def apply[V](thunk: => Cache[V]): Resource[F, Cache[V]] =
-      Resource.make(CatsAsync[F].delay(thunk))(toClose => Functor[F].map(toClose.close())(_ => ()))
-  }
+  def resourceCache[F[_]: CatsAsync: Functor: Mode, G[_], V](cacheF: F[G[V]])(
+      implicit ev: G[V] <~< Cache[V]): Resource[F, G[V]] =
+    Resource.make(cacheF)(toClose => Functor[F].map(ev(toClose).close())(_ => ()))
 
 }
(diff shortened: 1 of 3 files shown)


modules/cats-effect/src/main/scala/scalacache/CatsEffect.scala
```diff
@@ -1,20 +1,15 @@
...
 
-  object modes {
-
-    /**
-      * A mode that wraps computations in cats-effect IO.
-      */
-    implicit val io: Mode[IO] = new Mode[IO] {
-      val M: Async[IO] = asyncForCatsEffectAsync[IO]
-    }
+  object lowPrioModes {
```
cb372: I'm a bit confused. Why do you need to split this into 2 objects? This would be a breaking change for anyone who is using cats-effect `Async` but not `IO`.

GrafBlutwurst: oh yeah you're right. I was intending to make this a proper lowPrio trait but screwed it up. I had ambigous implicit issues for `Async[IO]` due to both the explicit instance `io` and the def `async` matching. Why are those both in there and not just the `async` one since that should cover cats `IO` as well?

cb372: Explanation of why both `IO` and `Async` are there: https://github.com/cb372/scalacache/pull/173#discussion_r182006753

But the `IO` one could probably be removed without anyone noticing, if it's causing you problems.

GrafBlutwurst:
So even though all the tests are passing now I have a weird situation provoking "diverging implicits"

```
import scalacache._
import scalacache.caffeine._
import scalacache.CatsEffect.modes._
import cats.effect.IO

object Main {

  def main(args:Array[String]) = {
    val cacheResource = CatsEffect.resourceCache(IO {CaffeineCache[String]})

    implicitly[Mode[IO]] //works
    implicitly[Flags] //works

    val prog = cacheResource.use(
      cache => for {
        //diverging implicits
        _ <- cache.put("1")("a")//(CatsEffect.modes.async[IO], Flags())
        a <- cache.get("1")//(CatsEffect.modes.async[IO], Flags())

      } yield a
    )


    println(prog.unsafeRunSync)

  }

}
```

GrafBlutwurst:
I fiddled around with the implicit resolution error but without success. So any help is welcome. I'll try to find a second pair of eyes to have a look at this.

lewisjkl:
Thinking that this will be done a different way now that #345 will remove modes and instead rely on cats-effect typeclasses. I am going to close this for now and we can look into more alternatives once #345 is merged in.

Issue #160, opened 2018-01-15
Support for Redis pipelining and multiple get requests


zainab-ali:
There are a couple of features which would be useful for REDIS.  In particular:
- getting and setting multiple keys in the same request
- pipelining (issuing multiple operations in the same request)

AFAIK, this is supported by jedis.  I'd be happy to have a go at adding these.

cb372:
Hi Zainab 👋 

Sure, I'd be happy to see a PR for this.

I'll be interested to see what kind of API you come up with for it, as I don't think any other backends support these features. ScalaCache often runs in to problems like this: how do you support the advanced features of some backends without having an adverse effect on people who want to use other backends?

By the way there is already an issue (#130) for getting multiple keys. Ignore my comment on that issue - I misunderstood and thought Sam wanted to get *every single item in the cache* in one call.

zainab-ali:
Yes, it's going to be tricky to implement.  Abstracting over all cache APIs gives the lowest common denominator, but that's not necessarily what would be most useful.  I'll see what I can come up with.

dmvieira:
I think it's an advanced usage of Scala Cache and by now, IMO, we could turns backend public to advanced functions usage. What do you think?

augi:
I'm willing to implement the multiple-key feature. I would implement it just for Redis now, so it would be implemented as new methods in `RedisCacheBase` class.

Would you mind accepting this kind of PR, @cb372, please?

Habitats:
@augi status on this? Would really love multi-key methods for put/get. Not having it is a major bottleneck for us.

augi:
We derived the RedisCache class in our project and added these methods there. 
I could prepare a new PR but I haven't got confirmation if a PR like this were accepted.

PR #253, opened 2019-04-26
withJedisCommands renamed to withJedis, and parameter changed to JClient

augi:
It supports more use-cases when used in derived classes. As the method is `protected` (like `JClient` type), it doesn't leaks anything to the public API.


cb372:
This looks reasonable. I'm curious about the use cases you have in mind. Is this related to your comment on #160?

augi:
Yes, it is. We have a class derived from your `RedisCache` that adds multi-key methods (and also methods to specify the key as `Array[Byte]`).

PR #257, opened 2019-05-06
Update com.twitter.util-core to 19.4.0

gurinderu:


build.sbt
```diff
@@ -44,6 +44,7 @@ lazy val root: Project = Project(id = "scalacache", base = file("."))
     catsEffect,
     monix,
     scalaz72,
+    twitterUtil,
```
cb372: Ah, so you noticed that this module was not aggregated and thus has never been published! I only spotted that the other day.

To be honest I'd like to slim down scalacache by removing some of the integration modules and replacing them with documentation and code samples. Some of these modules, including `twitter-util`, are literally one file, so they will be easy to turn into a code sample.

The reason for removing them is to make maintenance easier, specifically because I'd like to make a release for Scala 2.13 soon. `twitter-util` and `monix` are for the chop because they don't look like they will have Scala 2.13 support for a while.

How does this sound?

gurinderu: It wasn't aggregated because I forgot to add them before. 

About removing: I think it is better to have integrations for each library.
About 2.13: cross compilation should help)

lewisjkl:
Closing since `twitter-util` is no longer a dependency in the project.

PR #261, opened 2019-06-03
Support Redis Cluster

byamba-tumurkhuu:
Code duplication in `RedisCluster` from `RedisCacheBase` due to jedis cluster implementation being quite different from other jedis clients. It can be much simpler if https://github.com/xetorthio/jedis/issues/1545 was implemented. Adding wrappers around the java implementation might help. I did a little bit of that for the tests.

Please review: @cb372

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,10 @@ jdk:
 services:
   - memcached
   - redis
+install:
+  - make travis-install
+before_script:
+  - make travis-run
 script:
   - sbt scalafmtCheckAll scalafmtSbtCheck +coverage +test +doc/tut
   - sbt +coverageReport
@@ -18,3 +22,5 @@ cache:
 before_cache:
   - find $HOME/.sbt -name "*.lock" -type f -delete -print
   - find $HOME/.ivy2/cache -name "ivydata-*.properties" -type f -delete -print
+after_script:
+  - make stop
--- a/modules/redis/src/main/scala/scalacache/redis/RedisCluster.scala
+++ b/modules/redis/src/main/scala/scalacache/redis/RedisCluster.scala
@@ -1,0 +1,68 @@
+package scalacache.redis
+
+import redis.clients.jedis.JedisCluster
+import redis.clients.jedis.exceptions.JedisClusterException
+import scalacache.logging.Logger
+import scalacache.redis.StringEnrichment._
+import scalacache.serialization.Codec
+import scalacache.{AbstractCache, CacheConfig, Mode}
+
+import scala.concurrent.duration.{Duration, _}
+import scala.language.higherKinds
+
+class RedisCluster[V](val jedisCluster: JedisCluster)(implicit val config: CacheConfig, val codec: Codec[V])
+    extends AbstractCache[V] {
+
+  override protected final val logger = Logger.getLogger(getClass.getName)
+
+  override protected def doGet[F[_]](key: String)(implicit mode: Mode[F]): F[Option[V]] = mode.M.suspend {
+    val bytes = jedisCluster.get(key.utf8bytes)
+    val result: Codec.DecodingResult[Option[V]] = {
+      if (bytes != null)
+        codec.decode(bytes).right.map(Some(_))
+      else
+        Right(None)
+    }
+    result match {
+      case Left(e) =>
+        mode.M.raiseError(e)
+      case Right(maybeValue) =>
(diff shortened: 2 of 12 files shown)


byamba-tumurkhuu:
The tests for redis cluster weren't run since it's not setup on the ci server. It reduced the test coverage:
```
[error] (redis / coverageReport) Coverage minimum was not reached

```

byamba-tumurkhuu:
Trying to figure out how to add redis cluster to the ci build now.

byamba-tumurkhuu:
It looks like the redis version on travis CI is 4.0.6. `redis-cli` for 4.0.6 doesn't support creating redis cluster. I tried to compile the latest redis version unsuccessfully. I might have to just bring `redis-trib.rb` and try to create the cluster nodes.

.travis.yml
```diff
@@ -8,6 +8,10 @@ jdk:
 services:
   - memcached
   - redis
+install:
+  - make travis-install
+before_script:
+  - make travis-run
```
cb372: Well done for getting Redis Cluster working on Travis!

Maybe we could extend this in future to set up Redis Sentinel as well. Currently the Redis Sentinel tests are skipped on Travis.

modules/redis/src/main/scala/scalacache/redis/RedisCluster.scala
```diff
@@ -0,0 +1,68 @@
...
+
+  override protected def doRemove[F[_]](key: String)(implicit mode: Mode[F]): F[Any] = mode.M.delay {
+    jedisCluster.del(key.utf8bytes)
+  }
+
+  @deprecated(
+    "JedisCluster doesn't support this operation, scheduled to be removed with the next jedis major release",
+    "0.28.0"
+  )
+  override protected def doRemoveAll[F[_]]()(implicit mode: Mode[F]): F[Any] = mode.M.delay {
```
cb372: Instead of throwing an exception inside a `delay`, I would write this as

```scala
mode.M.raiseError(new JedisClusterException("No way to dispatch this command to Redis Cluster."))
```

cb372:
This looks great.

Regarding the duplication of code, could we work around the lack of shared interface by using a type class to represent the parts of the Jedis API we need? Although it's a bit tricky because it looks like `JedisCluster` doesn't use a `Pool` like the other clients do. Let me have a think about it.

modules/redis/src/main/scala/scalacache/redis/RedisCluster.scala
```diff
@@ -0,0 +1,68 @@
...
+import redis.clients.jedis.exceptions.JedisClusterException
+import scalacache.logging.Logger
+import scalacache.redis.StringEnrichment._
+import scalacache.serialization.Codec
+import scalacache.{AbstractCache, CacheConfig, Mode}
+
+import scala.concurrent.duration.{Duration, _}
+import scala.language.higherKinds
+
+class RedisCluster[V](val jedisCluster: JedisCluster)(implicit val config: CacheConfig, val codec: Codec[V])
```
cb372: Let's call this class `RedisClusterCache` for consistency with the existing classes.

modules/redis/src/main/scala/scalacache/redis/RedisCluster.scala
byamba-tumurkhuu: Done

.travis.yml
byamba-tumurkhuu: This is a bit hacky because it's using deprecated `redis-trib.rb` instead of `redis-cli`. `travis-install` can be removed if redis on travis is updated to v5. 

Yeah, this can totally be expanded to support Redis Sentinel.

byamba-tumurkhuu:
Yeah, It looked like it would require some major refactoring around `RedisCacheBase` in order to remove the code duplication. Perhaps that can be addressed with another PR? I'm happy to help as well.

cb372:
> Yeah, It looked like it would require some major refactoring around `RedisCacheBase` in order to remove the code duplication. Perhaps that can be addressed with another PR? I'm happy to help as well.

Yep, a separate PR sounds sensible. I'll merge this one now. Thanks!

byamba-tumurkhuu:
Thank you @cb372. When I do expect this change to be released?

cb372:
I'm waiting for a few dependencies to be released so I can do a release for Scala 2.13. Shouldn't be more than a few days.

PR #266, opened 2019-06-13
Add missing string interpolation to log

taylorwood:
Added string interpolation for `$d` parameter in new log warning from #261. This matches the equivalent log statements in other classes.


cb372:
Thanks!

PR #307, opened 2019-09-27
zio integration

dieproht:
Hey Chris, 
with this PR I made an integration for zio like there already exists for cats-effect, scalaz, try/future. Maybe you can have a look?
Karl

### DIFF
--- a/modules/zio/src/main/scala/scalacache/zio.scala
+++ b/modules/zio/src/main/scala/scalacache/zio.scala
@@ -1,0 +1,48 @@
+package scalacache
+
+import _root_.zio.Task
+
+import scala.util.control.NonFatal
+
+object zio {
+
+  object modes {
+
+    implicit val task: Mode[Task] = new Mode[Task] {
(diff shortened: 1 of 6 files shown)


modules/zio/src/main/scala/scalacache/zio.scala
```diff
@@ -0,0 +1,48 @@
+package scalacache
+
+import _root_.zio.Task
+
+import scala.util.control.NonFatal
+
+object zio {
```
dpennell: Can we have this as `ZioEffect` similar to `scalacache.CatsEffect`.
It is common to `import scalacache._` and this ends up shading the `zio` namespace.

dieproht: Thanks for your feedback. That sounds reasonable. I renamed the module object according your suggestion.

dpennell:
Thanks for contributing this!

lewisjkl:
Thanks for the contribution here! I am going to close this since the library as a whole is moving away from having different modes (see #345).

Issue #336, opened 2020-01-26
Let's save scalacache!


cb372:
I'd like to start with an apology. Over the last few months I have become increasingly apathetic towards maintaining this library. I haven't been merging scala-steward PRs, making new releases, or reviewing PRs from people who want to contribute. This is not fair on the library's users, and it's especially unfair on people who've given up their own time to make contributions.

The reasons for my neglect of the library are twofold. First is a simple lack of time and energy. I became a father almost 6 months ago, and since then my life has mostly revolved around my daughter. After doing my day job and looking after her, it's hard to summon the motivation to hack on OSS in my free time. I'm more inclined to veg out in front of Netflix for a couple of hours before bed. I'm acquainted with a few superhuman OSS contributors with kids, but I honestly don't know how they manage it.

But a much bigger reason for not wanting to interact with my own library is that I just don't like the way much of the code is written. I wrote most of it a long time ago and made a lot of bad design decisions along the way, and I have to confront my own mistakes every time I look at the code. It's frustrating to see these problems, knowing that I don't have enough free time to fix them.

So for a long time I've pretty much ignored all GitHub notifications for this repo, but without giving anyone the common courtesy of an explanation. That's not a reasonable or sustainable strategy, so the time has come to sort this mess out. In the remainder of this post I'd like to enumerate some of the problems I see with scalacache, and propose a plan to get the library into a shape where I would actually enjoy working on it.

If you'll indulge me in a little reminiscing, I'll start by summarising the history of the project in order to give a bit of context.

## Pre-history

The [very first commit](https://github.com/cb372/scalacache/commit/cbebdef3dd6bf911e2b04cdb467da369bc80cb88) was in February 2013, and the first release ([v0.1](https://github.com/cb372/scalacache/releases/tag/0.1)) was in November of that year. (I can't remember why it took so long to release.) Back then the library was called `cacheable`, and it was little more than the `memoize` macro. It was an experiment to see if I could use Scala macros (which were brand new at the time) to emulate the behaviour of Spring's `@Cacheable` annotation, auto-generating the cache key based on the method name and arguments.

The experiment was a success, and I felt the macro was useful, so I decided to expand it into a proper caching library so that people could actually make use of it. I didn't want to reinvent the wheel by writing my own cache implementation, so I made it an slf4j-style facade API, with integrations for a number of popular caching solutions. At the time there were integrations for Guava, Ehcache and Memcached. Later this grew to include Redis, Caffeine and others.

## v0.1 to 0.10

[...]

PR #345, opened 2020-02-07
Add cats-effect to the core module, remove modes

kubukoz:


modules/core/shared/src/main/scala/scalacache/AbstractCache.scala
```diff
@@ -13,34 +17,36 @@ import scala.language.higherKinds
   *
   * @tparam V The value of types stored in the cache.
   */
-trait AbstractCache[V] extends Cache[V] with LoggingSupport {
+trait AbstractCache[F[_], V] extends Cache[F, V] with LoggingSupport {
 
+  protected implicit def F: MonadError[F, Throwable]
+  protected implicit def Defer: Defer[F]
```
kubukoz: For now it looks like we can get away with just a dependency on cats-core.

kubukoz: At least until I noticed `Entry` and its timing method (which requires a clock).

lewisjkl:
@kubukoz this is looking great! Feels so much cleaner without the modes 😅.

cb372:
Great! The number of deleted lines is a good sign :) I’m on my phone now but I can take a proper look this weekend.

kubukoz:
@cb372 please do, if it looks good I'll proceed to migrate remaining modules.

modules/core/shared/src/main/scala/scalacache/AbstractCache.scala
cb372: Yeah. I think a cats-effect dependency in core is fine.

modules/core/shared/src/main/scala/scalacache/AbstractCache.scala
```diff
@@ -49,123 +55,92 @@ trait AbstractCache[V] extends Cache[V] with LoggingSupport {
       if (logger.isDebugEnabled) {
         logger.debug(s"Skipping cache PUT because cache writes are disabled. Key: $key")
       }
-      mode.M.pure(())
+      F.pure(())
```
cb372: `F.unit` ?

modules/core/shared/src/main/scala/scalacache/CacheAlg.scala
```diff
@@ -83,10 +73,8 @@ trait CacheAlg[V] {
     * It will take care of gracefully shutting down the underlying cache client.
     *
     * Note that you should not try to use this Cache instance after you have called this method.
-    *
-    * @param mode The operation mode, which decides the type of container in which to wrap the result
-    * @tparam F The type of container in which the result will be wrapped. This is decided by the mode.
     */
-  def close[F[_]]()(implicit mode: Mode[F]): F[Any]
+  //TODO: Replace with Resource-based API?
```
cb372: Possibly, but let's save that for another PR

lewisjkl:
> if it looks good I'll proceed to migrate remaining modules.

I know we talked in #336 about potentially removing some of the cache implementations. Could be worth drawing a box around which ones we'd want to remove (if any) prior to doing the work of actually migrating them. If so, I can open another issue so we can come to a decision about which ones to potentially remove.

modules/caffeine/src/main/scala/scalacache/caffeine/CaffeineCache.scala
```diff
@@ -1,79 +1,87 @@
...
-  def doGet[F[_]](key: String)(implicit mode: Mode[F]): F[Option[V]] = {
-    mode.M.delay {
-      val entry = underlying.getIfPresent(key)
-      val result = {
-        if (entry == null || entry.isExpired)
-          None
-        else
-          Some(entry.value)
+  def doGet(key: String): F[Option[V]] = {
+    F.delay {
```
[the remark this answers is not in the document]
kubukoz: I have a personal preference of putting by-name parameters in curly braces (if the parameter list only has one parameter), especially when lifting side effects. But I can adjust :)

modules/core/shared/src/main/scala/scalacache/CacheAlg.scala
kubukoz: sure

modules/core/shared/src/main/scala/scalacache/AbstractCache.scala
kubukoz: That'll need `.widen` - F is invariant and we need `F[Any]` unless I change the interface's type. Which I want to do, but maybe as a different change.

kubukoz:
@lewisjkl I'm totally on board with that, no point migrating code that is going to disappear. If you have some thoughts on which ones we'll remove, please bootstrap an issue :)

lewisjkl:
@kubukoz I just created #353 to discuss which cache implementations should be removed.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

cb372:
Hey @kubukoz thanks for waiting. We've now removed a few modules, so feel free to rebase and pick this work up again when you have time.

kubukoz:
Hey, I'll try to get back to this soon. Finished some other work so my bandwidth is a bit wider now...

modules/core/shared/src/main/scala/scalacache/AbstractCache.scala
kubukoz: Also, I was using Defer for lifting a (most-likely) side-effectful `=> V`, so Sync makes more sense anyway.

kubukoz:
@cb372 @lewisjkl how do you feel about logging? Currently we have impure logging all over the place, and it should be suspended.

I see the following options as best:

- keep the current logging but lift it manually via Sync
- depend on log4cats (it'll increase the amount of work to start using scalacache, as users will need to create a logger themselves)

For convenience, I'd go with the first, but not everyone will want to use slf4j (as we do now), which might be a problem later. What do you think?

lewisjkl:
> depend on log4cats (it'll increase the amount of work to start using scalacache, as users will need to create a logger themselves)

This would be great if not for requiring that extra setup. Do you think there’d be any good way to set a default to a no-op logger that is overridable if a user desires? If so, is that something worth doing?

Otherwise the sync option could be safer for now until we can think of a way to lower the friction to integrate and provide logging decoupled from a logging backend. I can definitely be persuaded here though.

kubukoz:
I think it's a reasonable course of action to proceed without anything smarter than Sync for now. Before the big bang release we can think about having it configurable. A no-op default certainly sounds possible in a way that makes it easy to override, as people don't create new caches all around the place (so less risk of inconsistent instances across the codebase).

lewisjkl:
@kubukoz is there anything that I can do to help push this along? I'm not trying to push you to do anything here, just know that I am happy to help if you don't currently have the bandwidth 😄

kubukoz:
> @kubukoz is there anything that I can do to help push this along? I'm not trying to push you to do anything here, just know that I am happy to help if you don't currently have the bandwidth 😄 

I'd love to finally do this, as it's been open for a while... I'll get back to it so I'm not blocking further improvements any longer.

cb372:
Yeah it would be great to get this one over the line 😄 Let me know if I can help.

kubukoz:
That should be it, for the most part... I'm sure I missed some `delay`s or forgot to put a `*>` between some effects (which used to be side effects), so I'll appreciate if you watch out for these during the review :)

kubukoz:
Apparently the redis cluster tests aren't working yet, and I haven't been able to set it up locally yet... I'll see what I can do.

modules/core/shared/src/main/scala/scalacache/Entry.scala
```diff
@@ -1,16 +1,31 @@
...
+    entry.expiresAt
+      .traverse { expiration =>
+        val now = clock.monotonic(TimeUnit.MILLISECONDS).map(Instant.ofEpochMilli(_))
 
+        now.map(expiration.isBefore(_))
+      }
+      .map {
+        case None | Some(true) => true
+        case Some(false)       => false
+      }
```
lewisjkl, line 30: What about `.forall(identity)`?
Maybe that would just be harder to read, but thought I'd throw it out there.

lewisjkl:
@kubukoz thanks for going through all all of that! Just looked through and didn't see any missed delays or `*>`. I'd be happy to take a look at the redis tests as well, but I don't want to step on your toes either. Just lmk if I can help 🙂

modules/core/shared/src/main/scala/scalacache/Entry.scala
kubukoz: Yeah, I had that but decided an explicit match would be more readable (in fact, I had a bug here due to having used `contains` / `exists`).

kubukoz:
> @kubukoz thanks for going through all all of that! Just looked through and didn't see any missed delays or `*>`. I'd be happy to take a look at the redis tests as well, but I don't want to step on your toes either. Just lmk if I can help 🙂 

I'll give it one more attempt today and let you know how it went :)

modules/core/shared/src/main/scala/scalacache/Entry.scala
lewisjkl: Cool, yeah I agree that the match is more readable. You don't have to do the mental gymnastics of remembering `forall` defaults to true.

modules/core/jvm/src/main/scala/scalacache/logging/Logger.scala
```diff
@@ -1,23 +1,26 @@
 package scalacache.logging
 
 import org.slf4j.{Logger => Slf4jLogger, LoggerFactory}
+import cats.effect.Sync
+import cats.Applicative
+import cats.implicits._
```
paulpdaniels, line 6: You don't appear to need this, based on what you are using I'd suggested targeted syntax:

```scala
import cats.syntax.option._
import cats.syntax.flatMap._
```

modules/core/shared/src/main/scala/scalacache/AbstractCache.scala
```diff
@@ -3,6 +3,11 @@ package scalacache
 import scala.concurrent.duration.Duration
 
 import scala.language.higherKinds
+import cats.Monad
+import cats.implicits._
```
paulpdaniels, line 7: targeted syntax for `functor`, `monadError` and `flatMap`

modules/core/shared/src/main/scala/scalacache/Cache.scala
```diff
@@ -3,19 +3,20 @@ package scalacache
...
+  def cachingForMemoize(baseKey: String)(ttl: Option[Duration])(
       f: => V
-  )(implicit mode: Mode[F], flags: Flags): F[V]
+  )(implicit flags: Flags): F[V]
 
-  def cachingForMemoizeF[F[_]](baseKey: String)(ttl: Option[Duration])(
-      f: => F[V]
-  )(implicit mode: Mode[F], flags: Flags): F[V]
+  def cachingForMemoizeF(baseKey: String)(ttl: Option[Duration])(
+      f: F[V]
```
paulpdaniels, line 19: Call by name might still be useful here, since I don't think there is a restriction against using `Future` or `Try` is there?

modules/core/shared/src/main/scala/scalacache/AbstractCache.scala
kubukoz: I'm not sure if there is a global standard for whether we want to use fine-grained imports or the whole thing. I can switch to either, but is there any improvement in using the targeted ones? The compilation time is roughly the same, from what I saw last time I checked.

modules/core/shared/src/main/scala/scalacache/Cache.scala
kubukoz: Neither Future nor Try have cats-effect instances, which you need anyway (Sync).

modules/core/shared/src/main/scala/scalacache/AbstractCache.scala
cb372: I try to use targeted imports as much as possible. Partially for compilation speed, although that is only anecdotal and wouldn't matter anyway on a project of this size. But more importantly for readability, as it gives readers more of a clue where instances and extension methods are coming from.

cb372:
Sorry everyone, I should have been pushing this PR a bit harder but I keep getting distracted.

@lewisjkl do you want to pick it up and try to get the tests working so we can get it merged and start working towards a release?

@kubukoz Thank you for all your work so far!

lewisjkl:
> @lewisjkl do you want to pick it up and try to get the tests working so we can get it merged and start working towards a release?

I would be happy to pick this up. I’ll take a look within the next couple of days and see if I can get those tests passing.

lewisjkl:
Started to look at the tests this morning. So far I tried to get a local redis cluster set up using docker-compose. I am able to get the tests in the `integrationtests` package to pass just fine, but the ones in `RedisClusterCacheSpec` are just hanging when I run them. I noticed that there is a make file in the project for setting up with Travis, but I am on a Mac rather than Linux so it didn't seem completely well suited for me. How do you typically go about running these locally @cb372? I am thinking I will keep going with the docker stuff I have up unless you have a better or more proven way.

cb372:
@lewisjkl I don't think I've ever run the cluster tests locally. Somebody else contributed that makefile and I've always relied on Travis to run those test for me.

In the past I installed redis via brew and fiddled around with config files to get Redis Sentinel working, but never tried for cluster. Docker sounds like a better idea.

lewisjkl:
@cb372 sounds good, thanks for the info. Mostly just wanted to make sure I wasn't reinventing the wheel here. I will see what I can get figured out with Docker. If I come up with something noteworthy then I will add it in a PR sometime soon.

lewisjkl:
So now all of the tests are passing and the docs are compiling. With the docs, I do think that we should go through and reformat some of the sections. Just to get them compiling, I added things like `.unsafeRunSync` when it is probably not preferable, but it will be a slightly larger rewrite to get them working using `.flatMap`/regions of sharing. Do you think I should take the time to do that on this PR? Or should we split that out so we can get this one merged?

I also noticed that there was a `todo` or two in the changeset for this PR. I will track them down and open individual comments about them in this thread so we can ensure they are addressed.

modules/core/shared/src/main/scala/scalacache/Cache.scala
```diff
@@ -3,19 +3,20 @@ package scalacache
 import scala.concurrent.duration.Duration
 import scala.language.higherKinds
 
-trait Cache[V] extends CacheAlg[V] {
+//todo merge with alg?
```
lewisjkl, line 6: Is this something we want to address in this PR? Seems like a good candidate for another PR to me, but happy to hear other opinions.

modules/core/shared/src/main/scala/scalacache/CacheAlg.scala
```diff
@@ -7,74 +7,64 @@ import scala.language.higherKinds
 /**
   * Abstract algebra describing the operations a cache can perform
   *
+  * @tparam F the effect of the cache. //todo elaborate
```
lewisjkl, line 10: I don't know how much elaboration is needed here. I think going too much more into depth on what `F[_]` means could be out of scope here.

modules/core/shared/src/main/scala/scalacache/package.scala
```diff
@@ -3,6 +3,7 @@ import scala.language.higherKinds
 
 package object scalacache {
 
+  //todo
```
lewisjkl, line 6: Is this todo indicating that `type Id[X] = X` is able to be removed? I am guessing that it can be removed in favor of people using cats Id when appropriate. Although it won't be usable for any of the algebras requiring Sync/Async anyway..

lewisjkl:
Also, for anyone curious, I ended up just using [this repo](https://github.com/carwestsam/redis-cluster-in-docker-osx) (I also removed the password that is set by default in this repo) to test with Redis locally.

modules/core/shared/src/main/scala/scalacache/CacheAlg.scala
cb372: Yeah I think we can just remove the `@tparam` line completely. We'll assume that anyone using the library is familiar with tagless-final.

modules/core/shared/src/main/scala/scalacache/package.scala
cb372: Yes I think this can be removed. I'll add a commit to do that.

cb372:
It finally happened! Merged! 🚀

kubukoz:
Wow, thanks a lot for making it happen @lewisjkl @cb372! I'm sorry I didn't get this to the end...

lewisjkl:
> Wow, thanks a lot for making it happen @lewisjkl @cb372! I'm sorry I didn't get this to the end...

No worries! Thanks for helping out so much with the refactor.

PR #347, opened 2020-02-11
Clarify Guava cache eviction in the docs.

stephennancekivell:
Hi, Thanks for the lib.

Can we add this note about eviction. I was a little surprised when I found this out.

Cheers


stephennancekivell:
Sorry, actually looking closer this isnt correct. It can evict old things on write.

PR #348, opened 2020-02-12
Multi operations in RedisCache

augi:
Partially solves #160. I couldn't implement these methods in the `RedisCacheBase` because `ShardedRedisCache` is not able to support the commands used.


ikryvorotenko:
any updates on this? Is this planned to be released anytime soon?

PR #352, opened 2020-02-16
Calculate TTL from value

agustafson:
I felt that it would make sense to calculate a TTL based on the value returned, rather than a hard-coded value. This still allows for hard-coded TTL calculation by using `(_ => ttl)`

As a result the order of parameters has changed:
`(f: => V)(calculateTtl: V => Option[Duration])`
ie the value is returned in the first param, and then the ttl is calculated from that value in the second param. This allows for type-inference to be used in the `calculateTtl` function. However, this means that it's a breaking change.

I've created `cachingForMemoizeLegacy` methods in the `Cache` trait, but haven't created a corresponding `memoization.legacy` package as I'm not sure whether the preference would be to support the old methods or force users to migrate to the new methods. I'm happy to add some support for legacy code if that is preferable.


kubukoz:
I'm not sure it's worth changing all the methods for this. Callers of the methods in question can calculate the TTL at call site (since you're asked to pass both a `V` and a duration), so we're not giving any extra power to the users.

agustafson:
> ~I'm not sure it's worth changing all the methods for this. Callers of the methods in question can calculate the TTL at call site (since you're asked to pass both a `V` and a duration), so we're not giving any extra power to the users.~
> 
> Disregard, that's completely untrue because of `=>` / `F` ;)

Yeah, in the situation that your underlying database (or whatever you're caching) returns an `F[_]` (which is almost always) then the changes will allow you to calculate a TTL from that.

kryptt:
I believe this functionality is worth the costs!

PR #363, opened 2020-03-03
Migrate to Mdoc

lewisjkl:
Working on moving from tut to mdoc. Sbt microsites has moved from tut to mdoc as the main typechecker.﻿


lewisjkl:
Looks like the site is rendering alright locally other than I don't see any styling show up locally. Not sure if that is normal so I will look into it more.

lewisjkl:
Well, I am not sure why, but I am seeing this:
image
Jekyll is reporting errors on loading css files. I am just running `jekyll serve -b /scalacache` from `target/site` after the site has been generated. Do you know why this might be happening @cb372?

cb372:
@lewisjkl Are you using an old Jekyll by any chance? Since 1.0 sbt-microsites requires Jekyll 4.x. I was caught out by that when I upgraded a different project.

lewisjkl:
Ah yeah I am on 3.x.x. I will upgrade that and see if it helps. Thanks for the input!

lewisjkl:
Okay, now on jekyll version 4.0.0 and the site generates just fine. Only thing now is I am not seeing the logo show up for some reason. Is that expected?

cb372:
Sorry, which logo? Scalacache doesn't have a logo so the site just shows the default sbt-microsites logo. If that's not showing up, I wouldn't worry about it. I assume it's just an artifact of running locally.

cb372:
Thanks for tackling this! A couple of comments:

* Let's move the files from `src/main/tut` to `src/main/mdoc` to avoid confusion in the future
* `.travis.yml` has a reference to `doc/tut` which will need to be updated to `doc/mdoc` (or maybe `doc/makeMicrosite` is a more useful test?)

lewisjkl:
> Scalacache doesn't have a logo
Gotcha. Yeah I think my confusion came from the default logo changing between different versions of sbt-microsites.

> move the files from src/main/tut to src/main/mdoc
This works for me, however I think the [default when using mdoc](https://47degrees.github.io/sbt-microsites/docs/getting-started/#write-your-documentation-md-files) with sbt-microsites is to just put the docs in `/docs`. Would we want to go that route? Or stick to `src/main/mdoc`?

> or maybe `doc/makeMicrosite` is a more useful test
I could see this being a good approach since that is what we actually run when generating the docs. I'd be good either way with this though.

cb372:
Cool, let's go with `docs/` if that's what sbt-microsites wants.

Let's update Travis to do `docs/makeMicrosite`.

lewisjkl:
I ended up going with the path of `modules/docs/src/main/mdoc` because I thought that was preferable to `modules/docs/src/main/docs`. Let me know if you see anything else that needs to be changed/improved.

cb372:
It looks like Travis reporting results to GitHub has been broken for days now, but the last build failed: https://travis-ci.org/github/cb372/scalacache/jobs/670194282#L409

`sbt docs/makeMicrosite` doesn't work because Jekyll isn't installed on Travis. Let's go with the easiest solution: just do `sbt docs/mdoc` instead.

I'll push a quick commit to your branch to fix that so we can get this merged.

PR #378, opened 2020-04-16
Zio integration

rleibman:
It is actually mostly @dieproht's changes from his branch, but I desperately need zio support. The only thing I changed from his was upgrading to the latest version of zio.


lewisjkl:
@rleibman we have been working on getting this library updated to use cats-effect typeclasses generically rather than trying to support cats-effect, scalaz, zio, etc at the same time. I am going to go ahead and close this PR since that is the direction that we are going and Zio has the ability to interop with those cats-effect typeclasses. But if I am missing something then please leave a comment here and we can reevaluate. Thanks!

rleibman:
I'm currently using my own fork, but I'd love official support for zio! Any timeline? 
Also... would you mind every once in a while showing up in gitter to let people know what's going on? I put in the PR back in April!

lewisjkl:
As far as a timeline goes, we need to get back in gear working toward another release. You can follow the discussion on this over on #410. Thank you for the feedback about being more active on the gitter channel. I think that would be good and I will try to be more active on this repo in general. We are trying to transition a few things over with this project (re: #410), but I hope once we have done that we can have a more regular release cycle.

PR #452, opened 2020-12-06
Remove Scala JS

lewisjkl:
A few main things went into removing Scala JS. First was getting rid of any plugins that were being used for JS support. Next was cleaning up directory structure since Scala JS introduced shared/, jvm/, and js/ in the core module. Finally, I went through the build.sbt file and modified any relevant places to remove JS support.


lewisjkl:
Looks like we are failing out on coverage percentage reached. Maybe because of the removal of Scala.js? Possible that it used to take some of that code into account when calculating the test coverage. I am thinking we should just lower the coverage limit for now since I didn't actually add any new code here. Thoughts @cb372?

For more context, it looks like we are getting `Coverage is below minimum [68.27% < 79.0%]` (see [here](https://travis-ci.org/github/cb372/scalacache/jobs/747894972#L849)).

kubukoz:
I don't believe in test coverage, personally I'd get rid of that check altogether..

lewisjkl:
After talking with @kubukoz about the test coverage, I am going to lower the required number to 60% for now to get this merged and we can revisit in the future.

PR #460, opened 2020-12-16
Upgrade to Cats Effect 3

### LINKED ISSUE #450: Update to cats-effect 3.0.0 milestones
Just making an issue to track the progress for https://github.com/typelevel/cats-effect/issues/1330.

Current milestone is 3.0.0-M4.

This should be an easy one - if I remember correctly, we're only using Sync/Clock, possibly some Async in a couple places. The migration should be quite seamless with just some renames or updated imports, but I might be missing something.

If anyone wants to take a stab at this, it should be a good first issue :) I'm happy to help in case of any trouble.

lewisjkl:
Main changes are:
  - supporting Clock w/o direct Applicative
  - updating async to async_
  - handling Sync usages that also required MonadError (now MonadCancel is needed in most cases)
  - import implicit unsafe IORuntime in tests
  - Update suspend to defer where applicable

Resolves #450

### DIFF
--- a/modules/caffeine/src/test/scala/scalacache/caffeine/CaffeineCacheSpec.scala
+++ b/modules/caffeine/src/test/scala/scalacache/caffeine/CaffeineCacheSpec.scala
@@ -7,15 +7,15 @@ import cats.effect.unsafe.{IORuntime, Scheduler}
 import cats.effect.{Clock, IO, Sync, SyncIO}
 import com.github.benmanes.caffeine.cache.Caffeine
 import org.scalatest.concurrent.ScalaFutures
-import org.scalatest.{AsyncFlatSpec, BeforeAndAfter, Matchers}
+import org.scalatest.{FlatSpec, BeforeAndAfter, Matchers}
 import scalacache._
 
 import scala.concurrent.ExecutionContext
 import scala.concurrent.duration._
 
-class CaffeineCacheSpec extends AsyncFlatSpec with Matchers with BeforeAndAfter with ScalaFutures {
+class CaffeineCacheSpec extends FlatSpec with Matchers with BeforeAndAfter with ScalaFutures {
 
-  private implicit val deterministicRuntime: (TestContext, IORuntime) = {
+  private val deterministicRuntime: (TestContext, IORuntime) = {
 
     val ctx = TestContext()
     val scheduler = new Scheduler {
@@ -39,7 +39,7 @@ class CaffeineCacheSpec extends AsyncFlatSpec with Matchers with BeforeAndAfter 
 
   private def newCCache = Caffeine.newBuilder.build[String, Entry[String]]
 
-  private def newFCache[F[_]: Sync: Clock, V](
+  private def newFCache[F[_]: Sync, V](
       underlying: com.github.benmanes.caffeine.cache.Cache[String, Entry[V]]
   ) = {
     CaffeineCache[F, V](underlying)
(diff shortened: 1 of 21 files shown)


lewisjkl:
Looks like I messed something up with timestamps in the caffeine cache. Going to take a look at this later today.

lewisjkl:
@kubukoz if you have a few minutes, I would love to have you look through this PR just to see if I am missing anything.

modules/caffeine/src/test/scala/scalacache/caffeine/CaffeineCacheSpec.scala
```diff
@@ -1,37 +1,54 @@
...
-import org.scalatest.concurrent.ScalaFutures
-import cats.effect.SyncIO
-import cats.effect.Clock
-import java.util.concurrent.TimeUnit
 
-class CaffeineCacheSpec extends FlatSpec with Matchers with BeforeAndAfter with ScalaFutures {
+class CaffeineCacheSpec extends AsyncFlatSpec with Matchers with BeforeAndAfter with ScalaFutures {
 
-  private def newCCache = Caffeine.newBuilder.build[String, Entry[String]]
+  private implicit val deterministicRuntime: (TestContext, IORuntime) = {
```
kubukoz, line 18: This probably doesn't need to be implicit

modules/caffeine/src/test/scala/scalacache/caffeine/CaffeineCacheSpec.scala
```diff
@@ -1,37 +1,54 @@
...
-  def newIOCache[V](
-      underlying: com.github.benmanes.caffeine.cache.Cache[String, Entry[V]],
-      clock: Clock[SyncIO] = defaultClock
+  private implicit val runTime: IORuntime   = deterministicRuntime._2
+  private val ctx: TestContext              = deterministicRuntime._1
+  private implicit val ec: ExecutionContext = ctx.derive()
+
+  private def newCCache = Caffeine.newBuilder.build[String, Entry[String]]
+
+  private def newFCache[F[_]: Sync: Clock, V](
```
kubukoz, line 42: Sync is already a clock, we can just leave Sync then

modules/caffeine/src/test/scala/scalacache/caffeine/CaffeineCacheSpec.scala
```diff
@@ -68,25 +86,28 @@ class CaffeineCacheSpec extends FlatSpec with Matchers with BeforeAndAfter with
...
-    underlying.getIfPresent("key1") should be(
-      Entry("hello", expiresAt = Some(Instant.parse("2015-10-31T00:00:00Z")))
-    )
+    val result = newFCache[IO, String](underlying).put("key1")("hello", Some(30.days)).unsafeToFuture().map { _ =>
+      underlying.getIfPresent("key1") should be(
+        Entry("hello", expiresAt = Some(now.plusMillis(30.days.toMillis)))
+      )
+    }
+    ctx.tick()
+    result
```
kubukoz, line 110: I think we can move the assertion here and ignore the result - basically do `.unsafeToFuture()`, forget about it, do `ctx.tick()` and make the assertion

kubukoz:
@djspiewak do you think TestContext should be able to create a Scheduler/IORuntime? I imagine many people will try to do that.

kubukoz (approved):

lewisjkl:
Thanks for the review @kubukoz. Just made those changes so once the pipeline passes I will get this merged.

kubukoz:
:shipit:

djspiewak:
@kubukoz Time mocking is one area that is a little weak in CE3 right now. In theory, `TimeT` was supposed to be the foundation of some new tooling there, but I haven't had the time to really flesh it out, and it has some spooky behavior with `race` that makes me nervous at present. Really it just needs some more thought and design.

Meanwhile, `TestContext` definitely *works*, but has some serious "handle with care" moments due to auto-yielding and such. I would rather not encourage people to use it, though it's hard not to given that creating a `Scheduler` from a `TestContext` is just about the only way to do this sort of thing at present. Also to make matters more complicated, there's no published artifact which includes *both* `Scheduler` and `TestContext`, so we would really need to spin up a new submodule. We might need to do that anyway, given that `Runners` is in the same boat and exposes some functionality which is really needed downstream.

kubukoz:
Now for M5! :D

PR #561, opened 2021-07-29
Cross-compile for Scala 3

hughsimpson:
Introduced a couple of little warts to retain identical keys for the tests (to avoid having to refactor them too much). Added comments on those locations.

Most of the changes are just formatting from bumping the scalafmt version...

Also (think I've) fixed the docs/mdoc call now 🤞 seems to work locally

Hopefully this is useful to someone else


modules/core/src/test/scala/scalacache/memoization/CacheKeySpecCommon.scala
```diff
@@ -50,7 +50,7 @@ trait CacheKeySpecCommon extends Suite with Matchers with BeforeAndAfter {
 
 }
 
-class AClass[F[_]](implicit cache: Cache[F, Int]) {
+class AClass[F[_]]()(implicit cache: Cache[F, Int]) {
```
hughsimpson, line 53: Added the extra `()` here just so they're included in the scala 3 generated cache keys (otherwise `CacheKeyIncludingConstructorParamsSpec."work for a method inside a class"` fails as it would generate `"scalacache.memoization.AClass(${cache.toString}).insideClass(1)"` rather than `"scalacache.memoization.AClass()(${cache.toString}).insideClass(1)"`)

modules/core/src/main/scala/scalacache/memoization/MethodCallToStringConverter.scala
```diff
@@ -33,7 +33,7 @@ object MethodCallToStringConverter {
 
   private def appendClassNamePart(sb: JStringBuilder)(className: String): Unit = {
     if (className.nonEmpty) {
-      sb.append(className)
+      sb.append(className.stripSuffix("$"))
```
hughsimpson, line 36: Ugly... Just to retain the exact same format for keys generated for the tests. Won't even do that in general, since an `object Foo { object Bar { class Baz }}` would have `Foo$.Bar.Baz` in scala 3 rather than `Foo.Bar.Baz` - so might be worth figuring out how to rewrite the test assertions appropriately to acknowledge the slightly different className formats between scala 2 and scala 3...

modules/core/src/main/scala/scalacache/memoization/MethodCallToStringConverter.scala
```diff
@@ -42,7 +42,7 @@ object MethodCallToStringConverter {
       sb: JStringBuilder
   )(className: String, constructorParamss: IndexedSeq[IndexedSeq[Any]]): Unit = {
     if (className.nonEmpty) {
-      sb.append(className)
+      sb.append(className.stripSuffix("$"))
```
hughsimpson, line 45: Ditto (see comment on the other `stripSuffix`)

modules/docs/src/main/mdoc/docs/memoization.md
```diff
@@ -99,7 +99,7 @@ class Bar(a: Int) {
 then you want the cache key to depend on the values of both `a` and `b`. In that case, you need to use a different implementation of [MethodCallToStringConverter](https://github.com/cb372/scalacache/blob/master/modules/core/shared/src/main/scala/scalacache/memoization/MethodCallToStringConverter.scala), like this:
 
 ```scala mdoc:silent
-implicit val cacheConfig = CacheConfig(
+implicit lazy val cacheConfig: CacheConfig = CacheConfig(
```
hughsimpson, line 102: Seems like the `implicit val catsCache: Cache[IO, Cat] = MemcachedCache("localhost:11211")` earlier on in this file constructs a forward reference to this implicit val (presumably it didn't before the tut->mdoc change); so this needs to be lazy or there's a weird NPE/AbstractMethodError (depending on scala version) thrown when running docs/mdoc

modules/core/src/main/scala-3/scalacache/memoization/Macros.scala
```diff
@@ -0,0 +1,82 @@
...
+
+  private def commonMacroImpl[F[_], V](
+      cache: Expr[Cache[F, V]],
+      keyNameToCachingCall: Expr[String] => Expr[F[V]]
+  )(using Quotes, Type[F], Type[V]): Expr[F[V]] = {
+    import quotes.reflect.*
+    val sym: Symbol = Symbol.spliceOwner
+
+    def hasCacheKeyExcludeAnnotation(s: Symbol): Boolean = s.annotations.exists {
+//      case Apply(Select(New(Ident("cacheKeyExclude")), _), _) => true // Don't know why this doesn't match...
```
kubukoz, line 33: funny, `summon[TypeTest[Tree, Ident]].unapply(o)` gives me a None too. Guess that's what's used, since `Ident` is a type member, thus prone to type erasure, so `TypeTest` will be used underneath (like `ClassTag` would be used in scala 2).

No idea why it doesn't match though. Maybe it's worth reporting somewhere...

modules/docs/src/main/mdoc/docs/memoization.md
kubukoz: Looks like the rest works without this snippet, so I'd rather remove the lazy but exclude this snippet from mdoc.

kubukoz (approved):

kubukoz:
This is good, thanks :)
updated with master and going to merge on green.

PR #562, opened 2021-07-29
Fix Expiration of Caffeine Cache

### LINKED ISSUE #522: Get returns nothing when TTL set
Hi,

It looks, that it's impossible to get value with `CaffeineCache`, if it was inserted with TTL.

Sample code:

```scala
import cats.effect.{ExitCode, IO, IOApp}
import scalacache.caffeine.CaffeineCache
import scalacache.{Cache, get, put}

object Main extends IOApp {
  val cache: IO[CaffeineCache[IO, String]] = CaffeineCache[IO, String]

  private def testCache(implicit c: Cache[IO, String]): IO[Unit] = for {
  _ <- put("foo")("bar", Some(scala.concurrent.duration.FiniteDuration(10, "days")))
  result <- get("foo")
  _ <- IO(println(result))
  } yield ()

  override def run(args: List[String]): IO[ExitCode] = cache.flatMap(implicit c => testCache.map(_ => ExitCode.Success))
}
```

It prints `None` but it should be `Some(bar)`. If I remove TTL, it works fine. My suspicion is that the following condition can be incorrect: https://github.com/cb372/scalacache/blob/1fb5879310a316553609ba4b026bb9089788eea7/modules/caffeine/src/main/scala/scalacache/caffeine/CaffeineCache.scala#L34

zarthross:
Fixes #522

### DIFF
--- a/modules/caffeine/src/main/scala/scalacache/caffeine/CaffeineCache.scala
+++ b/modules/caffeine/src/main/scala/scalacache/caffeine/CaffeineCache.scala
@@ -30,7 +30,7 @@ class CaffeineCache[F[_]: Sync, V](val underlying: CCache[String, Entry[V]])(imp
   def doGet(key: String): F[Option[V]] = {
     F.delay {
       Option(underlying.getIfPresent(key))
-    }.flatMap(_.filterA(Entry.isExpired[F, V]))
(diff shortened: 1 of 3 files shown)


kubukoz:
Thanks. I updated to master and made some simplifications to the super confusing code I wrote in the first place :D

Issue #517, opened 2021-04-21
1.0.0-M3 with CE3 final


kubukoz:
Seeing that cats-effect 3 is already included in the main branch, it would be nice to have a release ;)

If you need an extra pair of hands to push tags, I will gladly help every once in a while.

kubukoz:
1.0.0-M3 building now!

kubukoz:
aaaaaaand it failed due to the Tut artifacts not being available (bintray). I'll try to sort it out later.

kubukoz:
Released!

PR #565, opened 2021-08-13
Support polymorphic keys

ronnnnnnnnnnnnn:
With regard to standardizing the key serialization by making it part of AbstractCache - didn't think it was really necessary so preferred to keep it cleaner and more consistent with value serialization which is left up to the various implementations. I did make some changes to the codec library to allow for simple usage by implementations also for keys, and added such usage to the Redis implementation.
However, If you think it might still be better to standardize it some more, maybe I can add another layer for that inheriting from AbstractCache or something.
Anyway, I hope this is alright. Let me know what you think.


ronnnnnnnnnnnnn:
Merged the latest from master (including #517 and #561) and resolved the conflicts to make this compatible with the recent changes.

modules/caffeine/src/main/scala/scalacache/caffeine/CaffeineCache.scala
```diff
@@ -65,17 +60,17 @@ object CaffeineCache {
 
   /** Create a new Caffeine cache.
     */
-  def apply[F[_]: Sync: Clock, V](implicit config: CacheConfig): F[CaffeineCache[F, V]] =
-    Sync[F].delay(Caffeine.newBuilder().build[String, Entry[V]]()).map(apply(_))
+  def apply[F[_]: Sync: Clock, K <: AnyRef, V]: F[CaffeineCache[F, K, V]] =
```
lewisjkl, line 63: It looks like Caffeine requires the K and V to both extend `Object` (from Java). This wasn't a problem in the past because we were always using String keys. Weirdly, this also only seems to be an issue on Scala 2.12 as it compiles fine for 2.13 and 3. I had to add `<: AnyRef` just for 2.12, but I am not really sure of an alternative. @kubukoz Do you happen to have any thoughts on this?

lewisjkl: Added an issue #603 to discuss this further. Merging this PR so we can continue on with other things in the meantime.

lewisjkl:
Thanks again @ronnnnnnnnnnnnn for putting this together and sorry it took so long to get around to merging.

Issue #360, opened 2020-02-24
Upgrade to Jedis 3.x


cb372:
I haven't looked closely at the changes between 2.x and 3.x, but hopefully this wouldn't be too painful.

PR #580, opened 2021-09-08
upgrade jedis to 3.7.0

mayman:
Should address #360

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -67,7 +67,7 @@ lazy val memcached = createModule("memcached")
 lazy val redis = createModule("redis")
   .settings(
     libraryDependencies ++= Seq(
-      "redis.clients" % "jedis" % "2.10.2"
+      "redis.clients" % "jedis" % "3.7.0"
     ),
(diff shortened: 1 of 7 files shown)


mayman:
@kubukoz would you be able to have a look at this please? 

The upgrade to Jedis 3.7.0 allows the usage of SSL for redis, which is missing from the current implementation of jedis

kubukoz (approved):

kubukoz:
some compilation failures on 2.12 😅

mayman:
@kubukoz looks like I was missing some type info 🤦, I made sure to run the +test this time!

kubukoz:
Thanks @mayman :)

A snapshot should be released automatically to Sonatype in this job: https://github.com/cb372/scalacache/runs/3665273459?check_suite_focus=true
but I think we'll have a milestone release once we fix #522 (feel free to give it a shot if you want - info in comments on that issue). I think the current milestone is _less_ broken than the main branch so would like to avoid milestoning it.

cc @lewisjkl

PR #584, opened 2021-09-15
tidy up hasCacheKeyExcludeAnnotation

hughsimpson:
I figured out why the commented-out code didn't work before...


modules/core/src/main/scala-3/scalacache/memoization/Macros.scala
```diff
@@ -30,9 +30,8 @@ object Macros {
     val sym: Symbol = Symbol.spliceOwner
 
     def hasCacheKeyExcludeAnnotation(s: Symbol): Boolean = s.annotations.exists {
-//      case Apply(Select(New(Ident("cacheKeyExclude")), _), _) => true // Don't know why this doesn't match...
```
hughsimpson, line 33: This was nearly right, but despite the stringified form, the correct unapply method to call was `TypeIdent("cacheKeyExclude")` not `Ident("cacheKeyExclude")`

lewisjkl:
Awesome, thank you!

PR #591, opened 2021-09-23
fix cache entry expiration - monotinic vs realtime

### LINKED ISSUE #522: Get returns nothing when TTL set

yurique:
Fixes #522

### DIFF
--- a/modules/caffeine/src/test/scala/scalacache/caffeine/CaffeineCacheSpec.scala
+++ b/modules/caffeine/src/test/scala/scalacache/caffeine/CaffeineCacheSpec.scala
@@ -136,7 +136,7 @@ class CaffeineCacheSpec extends AnyFlatSpec with Matchers with BeforeAndAfter wi
       cache.get("key1").map { _ shouldBe defined }
   }
 
-  behavior of "get put with TTL"
+  behavior of "get after put with TTL"
 
   it should "store the given key-value pair with the given TTL, then get it back when not expired" in ticked {
     implicit ticker =>
--- a/modules/caffeine/src/test/scala/scalacache/caffeine/CaffeineCacheSpec.scala
+++ b/modules/caffeine/src/test/scala/scalacache/caffeine/CaffeineCacheSpec.scala
@@ -138,13 +138,33 @@ class CaffeineCacheSpec extends AnyFlatSpec with Matchers with BeforeAndAfter wi
 
   behavior of "get put with TTL"
 
-  it should "store the given key-value pair in the underlying cache with the given TTL, then get it back when not expired" in ticked {
+  it should "store the given key-value pair with the given TTL, then get it back when not expired" in ticked {
     implicit ticker =>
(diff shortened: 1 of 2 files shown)


yurique:
The tests make sense to me, but would be great if someone could double-check :)

kubukoz (approved):

modules/caffeine/src/test/scala/scalacache/caffeine/CaffeineCacheSpec.scala
```diff
@@ -114,4 +127,24 @@ class CaffeineCacheSpec extends AnyFlatSpec with Matchers with BeforeAndAfter wi
...
+  it should "store the given key-value pair in the underlying cache with no TTL, then get it back" in ticked { _ =>
+    val underlying = newCCache
+    val cache      = newIOCache(underlying)
+    cache.put("key1")("hello", None) *>
+      cache.get("key1").map { _ shouldBe defined }
+  }
+
+  behavior of "get put with TTL"
+
+  it should "store the given key-value pair in the underlying cache with the given TTL, then get it back when not expired" in ticked {
```
kubukoz, line 141: Let's have another one like this, but with an `IO.sleep` between put and get. It'll be instant because of how `ticked` works.

kubukoz:
Thanks! I always mix up the two kinds of clocks, this whole thing reminded me what the difference is 😅

modules/caffeine/src/test/scala/scalacache/caffeine/CaffeineCacheSpec.scala
yurique: Added a couple of tests with sleep.

kubukoz (approved):

kubukoz:
Think we could remove expired keys if we read one, as a space optimization - but that can be a different PR :)

PR #606, opened 2021-11-10
Cats-effect 3 typeclass coherence: remove clock from caffeine cache

zainab-ali:
Hi! 👋 

Thanks for all the wonderful work upgrading to cats effect 3. This is a very minor PR, but might help make the API a little cleaner.

The [typeclass hierarchy for cats-effect 3](https://typelevel.org/cats-effect/docs/typeclasses) has `Sync` extending `Clock`, meaning that the `Clock` typeclass is no longer necessary in the caffeine cache. 

Having the typeclass there causes user code to be prone to incoherence issues - a different `Clock` instance could be passed in along with the `Sync` instance, but the `Sync` instance would be favored in some cases of implicit resolution.


kubukoz (approved):

lewisjkl:
Thank you, this is a great find!

PR #665, opened 2022-03-24
Migrate project to use sbt-typelevel

DavidGregory084:
This migrates the project to use [sbt-typelevel](https://github.com/typelevel/sbt-typelevel). 

If it's decided to put the project forward as a Typelevel incubator project it might be good to do this anyway.

It also fixes the problems with doc generation that @lewisjkl saw since sbt-typelevel sets the `apiURL` for the project to ensure that cross-module API doc linking works.

A few other notable things:
* JDK17 was added to the build matrix as it's the new LTS
* Scala 2.13.x version was upgraded to 2.13.8
* The site is quite different now as it's generated via [Laika](https://planet42.github.io/Laika/index.html), and I think it will need theming appropriately as it's got Typelevel branding atm
* Outdated doc pages (relating to modes and the sync API) were removed

Contributed on behalf of the @opencastsoftware open source team. 👋


build.sbt
```diff
@@ -1,130 +1,156 @@
...
-    micrositeBaseUrl     := "/scalacache",
-    micrositeDocumentationUrl := "/scalacache/docs",
-    micrositeHomepage         := "https://github.com/cb372/scalacache",
-    micrositeGithubOwner      := "cb372",
-    micrositeGithubRepo       := "scalacache",
-    micrositeGitterChannel    := true,
-    micrositeTwitterCreator   := "@cbirchall",
-    micrositeShareOnSocial    := true,
-    mdocIn                    := (sourceDirectory in Compile).value / "mdoc"
+    tlSiteApiModule := Some((unidoc / projectID).value),
```
armanbilge, line 137: The `TypelevelUnidocPlugin` sets this one automatically 👍

DavidGregory084: Oh nice, also I had no idea about that `SettingKey` trick that you have used to do it, thanks!

build.sbt
```diff
@@ -12,7 +12,7 @@ val Slf4jVersion      = "1.7.26"
 
 lazy val scalatest     = "org.scalatest"     %% "scalatest"       % ScalaTestVersion         % Test
 lazy val scalacheck    = "org.scalacheck"    %% "scalacheck"      % "1.15.4"                 % Test
-lazy val scalatestplus = "org.scalatestplus" %% "scalacheck-1-15" % s"${ScalaTestVersion}.0" % Test
+lazy val scalatestplus = "org.scalatestplus" %% "scalacheck-1-15" % s"3.2.11.0" % Test // Cannot upgrade to 3.2.12.0 until Scalacheck 1.16.x
```
armanbilge, line 15: Hm, what's going on here?

DavidGregory084: Hmm perhaps we don't need to do this - I didn't realise that Scalacheck 1.16.x was binary compatible with 1.15.x

armanbilge: Scalacheck 1.16.x should be binary and source compatible with 1.15.x. So it should all work.

PR #889, opened 2025-04-24
Add unmaintenance notice

kubukoz:
I think it's best we manage the users' expectations - here's a README change, and afterwards I would archive the repository (it can still be unarchived, if need be)


kubukoz:
@cb372 pls, I really want to see this PR gone from my list 😿

